PR #15, opened 2017-07-07
Add support for formatting longs

sdorazio:
Currently, passing a `long` value as a bind parameter throws an exception. This will format the value the same way as it does `int` values.


laughingman7743:
Thanks!

PR #31, opened 2018-03-31
Cancel the query when KeybordInterrupt occurs (fix #28)

### LINKED ISSUE #28: Cancel aborted queries
After pressing Ctrl-C on `cur.execute` or `future.result` a `cur.cancel` should be sent to the server since athena is paid per query and there is no point in paying for the rest of an aborted query.

laughingman7743:


Spacerat:
I just tested this in a notebook and it seems to work just fine

laughingman7743:
https://github.com/laughingman7743/PyAthena/issues/28#issuecomment-377703455
> I don't think this is enough since the interruption could be catched at another part. A finally clause checking for the query execution state would be a more robust alternative, although that can't be done at the _poll level.

@Spacerat 
I think that test coverage is no problem. If this implementation looks good, I will merge this branch.

laughingman7743:
https://github.com/laughingman7743/PyAthena/pull/150

PR #34, opened 2018-04-22
Use pipenv & Update dependencies for test

laughingman7743:


greysteil:
Would you be up for using [Dependabot](https://dependabot.com) to automate creating dependency update PRs for your `Pipfile.lock`? I built it and it's totally free for open source - I'm just looking to get more feedback from Pipenv users. :octocat: (It might not handle this repo perfectly yet, but if it doesn't I'll get it fixed straight away.)

PR #58, opened 2018-12-18
Allow empty results from PandasCursor.

jkleint:
When running SQL statements without a result, such as DDL, PandasCursor would try to make a dataframe from an empty CSV file, which would crash.

This returns an empty DataFrame, allowing the use of DDL without error.


jkleint:
Ah, sorry about that.  Yes, your approach looks good.  Thanks!

PR #76, opened 2019-03-10
Update README

### LINKED ISSUE #60: PandasCursor doesn't automatically convert int columns with NA's to floats
I'm querying a large athena table and can successfully run a query using the below code, however it's really slow (for reasons covered in #46).

```python
conn = pyathena.connect(**at.athena_creds)
df = pd.read_sql(sql, conn)
```

I would really like to take advantage of the performance boost that PandasCursor offers, however, when I run the code below, I get a value error.

```python
conn = pyathena.connect(**at.athena_creds, cursor_class=PandasCursor)
cursor = at_con.cursor()
df = cursor.execute(sql).as_pandas()

>>> ValueError: Integer column has NA values in column 18
```

Now I understand why I'm getting this value error. I have a int column in my athena table which has NA values in it, which Pandas notoriously doesn't handle well (NaN's are floats in Pandas eyes, not ints). The `pd.read_sql()` seems to handle this gracefully. It recognizes there is an int column with NaN's and converts it to a float column. It would be great if pyathena did the same thing.

laughingman7743:
Add about ValueError of integer column in Dataframe. (close #60)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -435,10 +435,27 @@ Execution information of the query can also be retrieved.
     print(cursor.execution_time_in_millis)
     print(cursor.output_location)
 
+If the integer column has an NA value and ValueError occurs,
+use `Conversion Function`_ to convert the INTEGER column to a DOUBLE or FLOAT column.
+
+.. code:: python
+
+    from pyathena.connection import Connection
+    from pyathena.pandas_cursor import PandasCursor
+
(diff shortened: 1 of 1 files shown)


laughingman7743:
https://pandas.pydata.org/pandas-docs/stable/whatsnew/v0.24.0.html#optional-integer-na-support

PR #84, opened 2019-04-20
Impl AsyncPandasCursor (fix #81)

### LINKED ISSUE #81: Asynchronous Pandas?
It seems that one can either use the Async version of PyAthena, _or_ use the Pandas capabilities, either in the form of `pd.read_sql()` or in the form of `result_set.as_pandas()`.

It seems that using any form of Pandas integration makes all queries work in a synchronous manner, even if several calls are wrapped into a `ThreadPoolExecutor`.

Example:

```python
import pyathena
sql_engine = pyathena.connect(s3_staging_dir=app.config['S3_STAGING_DIR'])

def get_result(sql):
    print('sending sql_query at {}'.format(time.ctime(int(time.time()))))
    result = pd.read_sql(sql_query, sql_engine)
    return result

with concurrent.futures.ThreadPoolExecutor(max_workers=16) as executor:
    future_to_df = {executor.submit(get_result, item['sql']): item['measure_name'] for item in sql_queries}
    for future in concurrent.futures.as_completed(future_to_df):
        measure_name = future_to_df[future]
        raw_df_results[measure_name] = future.result()
```

Will output:

```
sending sql_query at Sun Apr 14 23:09:30 2019
sending sql_query at Sun Apr 14 23:09:30 2019
sending sql_query at Sun Apr 14 23:09:30 2019
```

But Athena's own history log will show that the queries were submitted synchronously:

```
# Athena history log
Submitted: 2019/04/15 00:09:30 UTC+-1 - Run time: 5s
Submitted: 2019/04/15 00:09:35 UTC+-1 - Run time: 6s
Submitted: 2019/04/15 00:09:41 UTC+-1 - Run time: 4s
```

Whereas if I use `pyathena.async_cursor.AsyncCursor` instead, Athena's logs will show that all queries were submitted at the same time, and therefore will treat them accordingly. However, there is, then, no way to use the Pandas integration from PyAthena.

What would be the best way to combine Pandas integration + Asynchronous submission?

laughingman7743:


laughingman7743:
TODO:
  - [ ] Update README

PR #95, opened 2019-10-08
WIP: Try to match previously executed queries

kbrose:
Add functionality that checks to see if the given query has been executed recently, and if so uses the results from that query (which Athena conveniently stores in S3). By default, this behavior is turned *off* to match the current behavior.

A new parameter `cache_size` is added to the `execute` method of `BaseCursor` subclasses. Only the previous `cache_size` queries are checked to see if they are a match.

A previously executed query is a "match" if it satisfies all of the following conditions:

* query strings match exactly
* the previous query succeeded
* the `s3` locations match
* previous query was a "DML" statement (so e.g. a `CREATE TABLE` statement will never be considered a "match")

Marked as Work In Progress until I can do some more testing.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -594,6 +594,24 @@ As with AsynchronousCursor, you need a query ID to cancel a query.
     query_id, future = cursor.execute("SELECT * FROM many_rows")
     cursor.cancel(query_id)
 
+Quickly re-run queries
+~~~~~~~~~~~~~~~~~~~~~~
+
+You can attempt to re-use the results from a previously run query to help save time and money in the cases where your underlying data isn't changing.
+
+.. code:: python
+
+    from pyathena import connect
+
+    cursor = connect(aws_access_key_id='YOUR_ACCESS_KEY_ID',
+                     aws_secret_access_key='YOUR_SECRET_ACCESS_KEY',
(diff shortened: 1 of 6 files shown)


kbrose:
@laughingman7743 it looks like the s3 test bucket is a hidden environment variable which doesn't get injected for PRs. What's the best way for me to test this change?

laughingman7743:
Secret environment variables cannot be used in PRs from forks.
https://docs.travis-ci.com/user/environment-variables/#defining-encrypted-variables-in-travisyml

You should set up TravisCI with your forked repository.
https://github.com/kbrose/PyAthena

The environment variables set in TravisCI are as follows.
```
AWS_ACCESS_KEY_ID
AWS_SECRET_ACCESS_KEY
AWS_ATHENA_S3_STAGING_DIR
AWS_DEFAULT_REGION
```

kbrose:
Ok, the only AWS account I have access to is the account I use at work, and I don't want to use that to run the tests if I can avoid it. Would you be opposed to adding my fork as a remote and checking out the `cache-results` branch and pushing it?

Let me know if that's too much.

kbrose:
Closing in favor of #96

PR #96, opened 2019-10-09
Cache results

kbrose:
Add functionality that checks to see if the given query has been executed recently, and if so uses the results from that query (which Athena conveniently stores in S3). By default, this behavior is turned off to match the current behavior.

A new parameter `cache_size` is added to the `execute` method of `BaseCursor` subclasses. Only the previous `cache_size` queries are checked to see if they are a match.

A previously executed query is a "match" if it satisfies all of the following conditions:

* query strings match exactly
* the previous query succeeded
* previous query was a "DML" statement (so e.g. a CREATE TABLE statement will never be considered a "match")

### DIFF
--- a/pyathena/common.py
+++ b/pyathena/common.py
@@ -142,28 +142,45 @@ class BaseCursor(with_metaclass(ABCMeta, object)):
             })
         return request
 
-    def _find_previous_query_id(self, request, work_group, cache_size):
-        next_token = None
-        while cache_size > 0:
-            n = min(cache_size, 50)  # 50 is max allowed by AWS API
-            cache_size -= n
-            request_params = {'MaxResults': n}
-            if work_group is not None:
-                request_params['WorkGroup'] = work_group
-            if next_token is not None:
-                request_params['NextToken'] = next_token
-            response = self.connection._client.list_query_executions(**request_params)
-            query_ids = response['QueryExecutionIds']
-            next_token = response['NextToken']
-            query_executions = self.connection._client.batch_get_query_execution(
-                QueryExecutionIds=query_ids
-            )['QueryExecutions']
-            for execution in query_executions:
-                queries_match = execution['Query'] == request['QueryString']
-                succeeded = execution['Status']['State'] == 'SUCCEEDED'
-                is_dml = execution['StatementType'] == 'DML'
-                if queries_match and succeeded and is_dml:
-                    return execution['QueryExecutionId']
+    def _find_previous_query_id(self, query, work_group, cache_size):
+        query_id = None
+        try:
+            next_token = None
+            while cache_size > 0:
+                n = min(cache_size, 50)  # 50 is max allowed by AWS API
+                cache_size -= n
+                request = {'MaxResults': n}
+                if work_group is not None:
+                    request['WorkGroup'] = work_group
+                if next_token is not None:
+                    request['NextToken'] = next_token
+                response = retry_api_call(self.connection._client.list_query_executions,
+                                          config=self._retry_config,
+                                          logger=_logger,
+                                          **request)
+                query_ids = response.get('QueryExecutionIds', None)
+                if not query_ids:
+                    break  # no queries left to check
+                next_token = response.get('NextToken', None)
+                query_executions = retry_api_call(
+                    self.connection._client.batch_get_query_execution,
+                    config=self._retry_config,
+                    logger=_logger,
+                    QueryExecutionIds=query_ids
+                ).get('QueryExecutions', [])
+                for execution in query_executions:
+                    if (
+                        execution['Query'] == query and
+                        execution['Status']['State'] == AthenaQueryExecution.STATE_SUCCEEDED and
+                        execution['StatementType'] == 'DML'
+                    ):
+                        query_id = execution['QueryExecutionId']
+                        break
+                if query_id or next_token is None:
+                    break
+        except Exception:
+            _logger.warning('Failed to check the cache. Moving on without cache.')
+        return query_id
 
     def _execute(self, operation, parameters=None, work_group=None, s3_staging_dir=None,
                  cache_size=0):
@@ -171,7 +188,7 @@ class BaseCursor(with_metaclass(ABCMeta, object)):
         _logger.debug(query)
 
         request = self._build_start_query_execution_request(query, work_group, s3_staging_dir)
-        query_id = self._find_previous_query_id(request, work_group, cache_size)
+        query_id = self._find_previous_query_id(request['QueryString'], work_group, cache_size)
         if query_id is None:
             try:
                 query_id = retry_api_call(self._connection.client.start_query_execution,
--- a/pyathena/common.py
+++ b/pyathena/common.py
@@ -142,24 +142,53 @@ class BaseCursor(with_metaclass(ABCMeta, object)):
             })
         return request
 
-    def _execute(self, operation, parameters=None, work_group=None, s3_staging_dir=None):
+    def _find_previous_query_id(self, request, work_group, cache_size):
+        next_token = None
+        while cache_size > 0:
+            n = min(cache_size, 50)  # 50 is max allowed by AWS API
+            request_params = {'MaxResults': n}
+            if work_group is not None:
+                request_params['WorkGroup'] = work_group
+            if next_token is not None:
+                request_params['NextToken'] = next_token
+            response = self.connection._client.list_query_executions(**request_params)
+            cache_size -= n
+            query_ids = response['QueryExecutionIds']
+            next_token = response['NextToken']
+            query_executions = self.connection._client.batch_get_query_execution(
+                QueryExecutionIds=query_ids
+            )['QueryExecutions']
+            requested_location = request['ResultConfiguration']['OutputLocation']
+            for execution in query_executions:
+                queries_match = execution['Query'] == request['QueryString']
+                succeeded = execution['Status']['State'] == 'SUCCEEDED'
+                executed_location = execution['ResultConfiguration']['OutputLocation']
+                locations_match = executed_location == requested_location
+                is_dml = execution['StatementType'] == 'DML'
+                if queries_match and succeeded and locations_match and is_dml:
+                    return execution['QueryExecutionId']
+
+    def _execute(self, operation, parameters=None, work_group=None, s3_staging_dir=None,
+                 cache_size=0):
         query = self._formatter.format(operation, parameters)
         _logger.debug(query)
 
         request = self._build_start_query_execution_request(query, work_group, s3_staging_dir)
-        try:
-            response = retry_api_call(self._connection.client.start_query_execution,
-                                      config=self._retry_config,
-                                      logger=_logger,
-                                      **request)
-        except Exception as e:
-            _logger.exception('Failed to execute query.')
-            raise_from(DatabaseError(*e.args), e)
-        else:
-            return response.get('QueryExecutionId', None)
+        query_id = self._find_previous_query_id(request, work_group, cache_size)
+        if query_id is None:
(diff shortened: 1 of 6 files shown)


kbrose:
@laughingman7743 I'm happy with how things look. Let me know what you think.

There seems to be a flaky test with the async pandas cursor (I'm fairly confident my changes aren't causing that test to fail) in the `test_cancel` function. Judging from https://travis-ci.org/laughingman7743/PyAthena/builds/567453898 it seems like this isn't the first time this test has caused problems.

pyathena/common.py
```diff
@@ -142,24 +142,53 @@ def _build_start_query_execution_request(self, query, work_group=None, s3_stagin
...
+            n = min(cache_size, 50)  # 50 is max allowed by AWS API
+            cache_size -= n
+            request_params = {'MaxResults': n}
+            if work_group is not None:
+                request_params['WorkGroup'] = work_group
+            if next_token is not None:
+                request_params['NextToken'] = next_token
+            response = self.connection._client.list_query_executions(**request_params)
+            query_ids = response['QueryExecutionIds']
+            next_token = response['NextToken']
```
laughingman7743: If the cache size is set to 1000 or larger, a large number of requests will be thrown even if there is almost no query execution history.
This loop should be stopped when the next_token of the response from the API is None.

pyathena/common.py
```diff
@@ -142,24 +142,53 @@ def _build_start_query_execution_request(self, query, work_group=None, s3_stagin
             })
         return request
 
-    def _execute(self, operation, parameters=None, work_group=None, s3_staging_dir=None):
+    def _find_previous_query_id(self, request, work_group, cache_size):
```
laughingman7743: The signature of this method is as follows:
```def _find_previous_query_id (self, query, s3_staging_dir, work_group, cache_size):```
I want the `request_params` variable to be renamed to `request`.

File: pyathena/common.py [hunk 3]
```diff
@@ -142,24 +142,53 @@ def _build_start_query_execution_request(self, query, work_group=None, s3_stagin
...
+        next_token = None
+        while cache_size > 0:
+            n = min(cache_size, 50)  # 50 is max allowed by AWS API
+            cache_size -= n
+            request_params = {'MaxResults': n}
+            if work_group is not None:
+                request_params['WorkGroup'] = work_group
+            if next_token is not None:
+                request_params['NextToken'] = next_token
+            response = self.connection._client.list_query_executions(**request_params)
```
laughingman7743: Use `retry_api_call` so that API requests can be retried.

pyathena/common.py
```diff
@@ -142,24 +142,53 @@ def _build_start_query_execution_request(self, query, work_group=None, s3_stagin
...
-                                      logger=_logger,
-                                      **request)
-        except Exception as e:
-            _logger.exception('Failed to execute query.')
-            raise_from(DatabaseError(*e.args), e)
-        else:
-            return response.get('QueryExecutionId', None)
+            query_id = self._find_previous_query_id(request, work_group, cache_size)
+        except Exception:
+            _logger.warn('Failed to check the cache. Moving on without cache.')
```
laughingman7743: I think that the process of receiving an exception here is not appropriate.
I think it should be supported by the `_find_previous_query_id` method.

laughingman7743:
@kbrose 
I think it ’s a good improvement. 👍 
If you set cache_size, the behavior should not change.
Please check out some comments.

kbrose:
Thank you for reviewing! I tried to address the comments in the most recent commit.

pyathena/common.py
```diff
@@ -142,24 +142,59 @@ def _build_start_query_execution_request(self, query, work_group=None, s3_stagin
...
+                next_token = response.get('NextToken')
+                if next_token is None:
+                    # No more queries in Athena history. Stop iterating.
+                    cache_size = 0
+                query_executions = self.connection._client.batch_get_query_execution(
+                    QueryExecutionIds=query_ids
+                )['QueryExecutions']
+                for execution in query_executions:
+                    queries_match = execution['Query'] == query
+                    succeeded = execution['Status']['State'] == 'SUCCEEDED'
```
laughingman7743: Please use `AthenaQueryExecution.STATE_SUCCEEDED`.

tests/test_cursor.py
```diff
@@ -71,6 +71,29 @@ def test_iterator(self, cursor):
         self.assertEqual(list(cursor), [(1,)])
         self.assertRaises(StopIteration, cursor.__next__)
 
+    @with_cursor
+    def test_cache_size(self, cursor):
+        # To test cacheing, we need to make sure the query is unique, otherwise
+        # we might accidentally pick up the cache results from another CI run.
+        unique_comment = "-- " + str(datetime.utcnow())
```
laughingman7743: I thought it would be better to define the query variable as follows.
query = `SELECT * FROM one_row -- {0}`.format(str(datetime.utcnow()))

pyathena/common.py
```diff
@@ -142,24 +142,59 @@ def _build_start_query_execution_request(self, query, work_group=None, s3_stagin
...
+                if work_group is not None:
+                    request['WorkGroup'] = work_group
+                if next_token is not None:
+                    request['NextToken'] = next_token
+                response = retry_api_call(self.connection._client.list_query_executions,
+                                          config=self._retry_config,
+                                          logger=_logger,
+                                          **request)
+                query_ids = response['QueryExecutionIds']
+                next_token = response.get('NextToken')
```
laughingman7743: I prefer to write explicit default values.
`next_token = response.get('NextToken', None)`

kbrose:
Addressed further comments

pyathena/common.py
```diff
@@ -142,24 +142,64 @@ def _build_start_query_execution_request(self, query, work_group=None, s3_stagin
             })
         return request
 
-    def _execute(self, operation, parameters=None, work_group=None, s3_staging_dir=None):
+    def _find_previous_query_id(self, query, work_group, cache_size):
```
laughingman7743: This method may return a QueryID or nothing. If possible, I'd like to return QueryID or None explicitly.
For example, the following implementation:
```
    def _find_previous_query_id(self, query, work_group, cache_size):
        query_id = None
        try:
            next_token = None
            while cache_size > 0:
                n = min(cache_size, 50)  # 50 is max allowed by AWS API
                cache_size -= n
                request = {'MaxResults': n}
                if work_group is not None:
                    request['WorkGroup'] = work_group
                if next_token is not None:
                    request['NextToken'] = next_token
                response = retry_api_call(self.connection._client.list_query_executions,
                                          config=self._retry_config,
                                          logger=_logger,
                                          **request)
                query_ids = response.get('QueryExecutionIds', None)
                if not query_ids:
                    break  # no queries left to check
                next_token = response.get('NextToken', None)
                query_executions = retry_api_call(
                    self.connection._client.batch_get_query_execution,
                    config=self._retry_config,
                    logger=_logger,
                    QueryExecutionIds=query_ids
                ).get('QueryExecutions', [])
                for execution in query_executions:
                    if execution['Query'] == query and \
                            execution['Status']['State'] == AthenaQueryExecution.STATE_SUCCEEDED and \
                            execution['StatementType'] == 'DML':
                        query_id = execution['QueryExecutionId']
                        break
                if query_id or next_token is None:
                    break
        except Exception:
            _logger.warning('Failed to check the cache. Moving on without cache.')
        return query_id
```

pyathena/common.py
```diff
@@ -142,24 +142,64 @@ def _build_start_query_execution_request(self, query, work_group=None, s3_stagin
...
+                    queries_match = execution['Query'] == query
+                    success = execution['Status']['State'] == AthenaQueryExecution.STATE_SUCCEEDED
+                    is_dml = execution['StatementType'] == 'DML'
+                    if queries_match and success and is_dml:
+                        return execution['QueryExecutionId']
+                if next_token is None:
+                    # no queries left to check
+                    break
+        except Exception:
+            _logger.warn('Failed to check the cache. Moving on without cache.')
```
laughingman7743: "The 'warn' method is deprecated, use 'warning' instead.

pyathena/common.py
```diff
@@ -142,24 +142,64 @@ def _build_start_query_execution_request(self, query, work_group=None, s3_stagin
...
+                query_executions = retry_api_call(
+                    self.connection._client.batch_get_query_execution,
+                    config=self._retry_config,
+                    logger=_logger,
+                    QueryExecutionIds=query_ids
+                ).get('QueryExecutions', [])
+                for execution in query_executions:
+                    queries_match = execution['Query'] == query
+                    success = execution['Status']['State'] == AthenaQueryExecution.STATE_SUCCEEDED
+                    is_dml = execution['StatementType'] == 'DML'
```
laughingman7743: I thought there was no reason to assign it to a variable.

kbrose: I find it more readable than a multi-line if statement, but 🤷‍♂

pyathena/common.py
kbrose: Won't this fail your linting which restricts lines to <= 100 characters?

kbrose: Yep, it fails the linting. https://travis-ci.org/laughingman7743/PyAthena/jobs/596109171#L289

kbrose:
Added more commits addressing comments

laughingman7743:
Thanks you.
Would you like to squash this branch commit?

kbrose:
Squashing is fine. Do you mean you want me to rebase my commits into one commit? Or that when you click the merge button you'll choose the squash option?

laughingman7743:
I don't like the squash when merging this branch to the master because it doesn't seem to respect the contributors' intentions.
I think that it is not necessary to make one commit if it is a meaningful commit.

kbrose:
Ok, I'm not quite sure I follow, but please let me know if there is more you want me to do for this PR.

laughingman7743:
I don't like commit logs like "PR comments" or "more comments".
I'd like you to fix it to a meaningful commit log or squash to a single commit.
It's okay to force push to this branch.

kbrose:
Reworded/squashed some commits so it should be more legible now. You can reference https://github.com/kbrose/PyAthena/commits/cache-results to see what the commits looked like before the rebase.

I believe the following verifies that I didn't mess things up in the rebase.

```
$ git remote -v
origin	git@github.com:kbrose/PyAthena.git (fetch)
origin	git@github.com:kbrose/PyAthena.git (push)
upstream	git@github.com:laughingman7743/PyAthena.git (fetch)
upstream	git@github.com:laughingman7743/PyAthena.git (push)
$ git diff upstream/cache-results origin/cache-results | wc -l
0
```

Let me know what you think.

laughingman7743:
Which is better for the next version, 1.7.3 or 1.8.0? 🤔 
Do you want to write a release description?

kbrose:
Thank you, both for creating this great library and for working through the review process with me!

I'd say 1.8.0, I would think of 1.7.3 as a bug-fix release. But I don't think it matters much either way (to me at least).

For the release description, you could use what I wrote in the README section if you think that was informative:

> You can attempt to re-use the results from a previously run query to help save time and money in the cases where your underlying data isn't changing. Set the `cache_size` parameter of `cursor.execute()` to a number larger than 0 to enable cacheing. (Cacheing is turned *off* by default.) See the [README](https://github.com/laughingman7743/PyAthena/blob/master/README.rst#quickly-re-run-queries) for details.

laughingman7743:
I fixed some points I was interested in.
Please do not worry. You did a great job. 👍 

v1.8.0 released to PyPI. It can be installed with Pip. 🎉 
https://github.com/laughingman7743/PyAthena/releases/tag/v1.8.0
https://pypi.org/project/PyAthena/1.8.0/

PR #104, opened 2019-11-23
Support NA values with boolean column (fix #100, fix #102, fix #103)

### LINKED ISSUE #100: Error handling for boolean columns with nulls/missing values using as_pandas() from PandasCursor
I'm getting an error: `Bool column has NA values in column 31` for cases where I'm trying to do a quick `SELECT * FROM table LIMIT 5` type of query when there are boolean columns containing missing values. I was wondering if there was any way to build in handling for this that would allow me to still pull in the data into a DataFrame for these cases.

### LINKED ISSUE #102: Support missing values in Boolean columns
https://github.com/laughingman7743/PyAthena/pull/80
Similar to the above issue, I ran into an issue with missing values in Boolean Columns. When parsing to pandas the conversion fails because we set the dtype to boolean while boolean does not support Na. 

```
pandas/_libs/parsers.pyx in pandas._libs.parsers.TextReader._read_rows()
pandas/_libs/parsers.pyx in pandas._libs.parsers.TextReader._convert_column_data()
pandas/_libs/parsers.pyx in pandas._libs.parsers.TextReader._convert_tokens()
pandas/_libs/parsers.pyx in pandas._libs.parsers.TextReader._convert_with_dtype()
ValueError: Bool column has NA values in column 3
```
Could we add a try except Value Error around https://github.com/laughingman7743/PyAthena/blob/d20213533387d473386e32d49bba1cf8616ed253/pyathena/result_set.py#L429 
If it fails we could raise a warning and try again the DataFrame creation with relaxed constraints: 
- Either we follow pandas and convert to object the column involved in the issue (I'm not sure we can detect the faulty column, so we could convert all columns with dtypes that are not Na compatible.)
- Convert booleans (I understand from the PR above that int is already na safe ) to int and warn the user.

If needed I can write a PR once we agree on a solution ! 

Thanks

laughingman7743:


tests/test_util.py
```diff
@@ -0,0 +1,122 @@
...
+        """)
+        df = as_pandas(cursor)
+        rows = [tuple([
+            row['a'],
+            row['b'],
+        ]) for _, row in df.iterrows()]
+        self.assertEqual(rows, [
+            (True, False),
+            (False, None),
+            (None, None),
```
tanguycdls: 👍 Nice tests ! I understand that since you dont provide a dtype for the boolean column when calling read_csv. Pandas cast it either to boolean if there are no missing values or object if there are some ?

That would explain why it's a None and not a na. 

Thanks a lot for the prompt PR :)

laughingman7743: I wrote how to customize dtypes and converters in README.

laughingman7743: https://github.com/laughingman7743/PyAthena/blob/6276338ed0b10df8625ddc3090fa099e3ab9b5da/pyathena/converter.py#L145
By default this class is used.

PR #127, opened 2020-03-21
Changed date and datetime type conversions to use literal values in all cases (fix #126)

### LINKED ISSUE #126: Date(time) parameters lose type in CTAS
Consider the query:

```
import datetime
import pyathena
cursor = pyathena.connect().cursor()
cursor.execute('CREATE TABLE test AS SELECT %(date)s as date', {'date': datetime.date(2018,1,1)})
cursor.execute('SELECT * FROM test')
result = cursor.fetchone()
```

Expected output:
```
result == (datetime.date(2018,1,1),)
```

Actual output:
```
result == '2018-01-01'
```

Explanation:
In `formatter.py` we have

    def _format_date(formatter, escaper, val):
    if escaper is _escape_presto:
        return "date'{0}'".format(val.strftime("%Y-%m-%d"))
    else:
        return "'{0}'".format(val.strftime("%Y-%m-%d"))

and the escaper is set in `DefaultParameterFormatter`:

        if operation.upper().startswith('SELECT') or operation.upper().startswith('WITH'):
            escaper = _escape_presto
        else:
            escaper = _escape_hive

So in this case because our operation starts with `CREATE` it does an `_escape_hive` and removes the date format. (Similary for datetime)

Is there a reason this case needs to be treated differently?

laughingman7743:
TODO:
- [x] Unit test.

### DIFF
--- a/pyathena/formatter.py
+++ b/pyathena/formatter.py
@@ -44,17 +44,11 @@ def _format_default(formatter, escaper, val):
 
 
 def _format_date(formatter, escaper, val):
-    if escaper is _escape_presto:
-        return "date'{0}'".format(val.strftime("%Y-%m-%d"))
-    else:
-        return "'{0}'".format(val.strftime("%Y-%m-%d"))
+    return "CAST('{0}' AS DATE)".format(val.strftime("%Y-%m-%d"))
 
 
 def _format_datetime(formatter, escaper, val):
-    if escaper is _escape_presto:
-        return "timestamp'{0}'".format(val.strftime('%Y-%m-%d %H:%M:%S.%f')[:-3])
-    else:
-        return "'{0}'".format(val.strftime('%Y-%m-%d %H:%M:%S.%f')[:-3])
+    return "CAST('{0}' AS TIMESTAMP)".format(val.strftime('%Y-%m-%d %H:%M:%S.%f')[:-3])
 
 
 def _format_bool(formatter, escaper, val):
--- a/pyathena/formatter.py
+++ b/pyathena/formatter.py
@@ -44,11 +44,11 @@ def _format_default(formatter, escaper, val):
 
 
 def _format_date(formatter, escaper, val):
-    return "CAST('{0}' AS DATE)".format(val.strftime("%Y-%m-%d"))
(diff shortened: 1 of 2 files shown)


laughingman7743:
```
_______________ TestDefaultParameterFormatter.test_add_partition _______________
self = <tests.test_formatter.TestDefaultParameterFormatter testMethod=test_add_partition>
    def test_add_partition(self):
        expected = """
        ALTER TABLE test_table
        ADD PARTITION (dt='2017-01-01', hour=1)
        """.strip()

        actual = self.format("""
        ALTER TABLE test_table
        ADD PARTITION (dt=%(dt)s, hour=%(hour)d)
        """, {'dt': date(2017, 1, 1), 'hour': 1})
>       self.assertEqual(actual, expected)
E       AssertionError: "ALTE[19 chars]n        ADD PARTITION (dt=CAST('2017-01-01' AS DATE), hour=1)" != "ALTE[19 chars]n        ADD PARTITION (dt='2017-01-01', hour=1)"
E         ALTER TABLE test_table
E       -         ADD PARTITION (dt=CAST('2017-01-01' AS DATE), hour=1)?                           -----            ---------
E       +         ADD PARTITION (dt='2017-01-01', hour=1)
```

laughingman7743:
```
CREATE EXTERNAL TABLE IF NOT EXISTS date_partition_table (
    a STRING
)
PARTITIONED BY (b DATE)
ROW FORMAT DELIMITED FIELDS TERMINATED BY '\t' LINES TERMINATED BY '\n' STORED AS TEXTFILE
LOCATION 's3://BUCKET/path/to/';
```

A partition can be added with the following ALTER statement.
`ALTER table date_partition_table ADD PARTITION (b=DATE'2017-01-01');`

An error occurred in the following ALTER statement.
`ALTER table date_partition_table ADD PARTITION (b=CAST('2017-01-01' AS DATE));`
`line 1:38: missing 'column' at 'partition' (service: amazonathena; status code: 400; error code: invalidrequestexception;`

laughingman7743:
In all cases, it may be better to convert to DATE'2020-01-01 'or TIMESTAMP'2020-01-01 00:00:00.000'.

PR #182, opened 2020-11-28
Update README.rst

l1x:
Replacing putting credentials into code / version control with referencing .aws/credentials profiles. Furthermore we could add documentation for using instance profile credentials or boto3 session credentials for using PyAthena with AssumeRole credentials.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -58,8 +58,7 @@ Basic usage
 
     from pyathena import connect
 
-    cursor = connect(aws_access_key_id="YOUR_ACCESS_KEY_ID",
-                     aws_secret_access_key="YOUR_SECRET_ACCESS_KEY",
+    cursor = connect(profile_name='aws-profile-name',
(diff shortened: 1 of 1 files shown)


laughingman7743:
You don't even need to specify the profile_name if the default profile settings are in place.
The README describes the most straightforward way to connect to Athena, which is why we explicitly specify AWS credentials.
If you want to improve the documentation, it would be nice to create a section on how to specify the profile name.

l1x:
> The README describes the most straightforward way to connect to Athena, which is why we explicitly specify AWS credentials.

What you classify as  the most straightforward  is actively discouraged by the vendor.

- https://docs.aws.amazon.com/general/latest/gr/aws-access-keys-best-practices.html

> Don't embed access keys directly into code. The AWS SDKs and the AWS Command Line Tools enable you to put access keys in known locations so that you do not have to keep them in code.

I don't think that there is a significant user experience difference between aws profiles vs access key / secret key but there is a huge security difference.

PR #221, opened 2021-03-13
Support for catalog (fix #220) & Change schema_name to optional

### LINKED ISSUE #220: Support for Catalog
I have noticed that PyAthena currently assumes only `Database` is enough as part of the [StartQueryExecution API call](https://docs.aws.amazon.com/athena/latest/APIReference/API_StartQueryExecution.html): https://github.com/laughingman7743/PyAthena/blob/e4a74ddfd7d976a968e18e83d89487f020ec48a6/pyathena/common.py#L185

Would it be possible to consider adding `Catalog` as well in case non-default data source (`AwsDataCatalog`) is used?

laughingman7743:


milanaleksic:
@laughingman7743 thank you for your quick feedback and the work! I have tried out this branch, and indeed I see the `Catalog` property being sent to AWS so as far as I am concerned the issue is solved!

PR #252, opened 2021-11-21
Allow the boto3 session object to pass as an argument to the connection method (fix #248)

### LINKED ISSUE #248: boto3 session as argument to connect()
Currently, you can provide arguments that are used to create a boto3 session, but you can't just provide your own session. That limits the ability, for example, to use [`aws_sso_lib.get_boto3_session()`](https://github.com/benkehoe/aws-sso-util/tree/master/lib#get_boto3_session) to allow a script that knows what account and AWS SSO role it should be using to do that directly, rather than relying on it being configured in a profile.

laughingman7743:


pyathena/connection.py
```diff
@@ -66,6 +66,7 @@ def __init__(
         retry_config: Optional[RetryConfig] = None,
         cursor_class: Type[BaseCursor] = Cursor,
         kill_on_interrupt: bool = True,
+        session: Any = None,
```
benkehoe, line 69: This could use `Optional[Session]` rather than `Any`

laughingman7743: Fixed: https://github.com/laughingman7743/PyAthena/pull/252/commits/8603a2d97b80a75bc48ea463ace3b18efef6a989

PR #255, opened 2021-12-01
[pandas] - Support partitions with special chars table names

fernbach:
current implementation doesn't support add partition (MSCK REPAIR TABLE) for table names / schema names that starts with number / underscore.

this simple change fix this bug


laughingman7743 (approved):
Thank you!

Issue #263, opened 2022-01-15
Add support for table and column comments.


cansjt:
Though Athena supports them, as of today, there are no ways to set comments in a table definition. Neither on the the table itself or on the table's columns.

We'd like to be capable of doing that, with SQLAlchemy, that defines the interfaces needed to implement this.

Example:
```python
from sqlalchemy.sql.ddl import CreateTable
from sqlachemy import Column, Integer, MetaData, Table

from pyathena.sqlachemy_athena import AthenaDialect


table = Table('table_name',
              MetaData(),
              Column('c', Integer, comment='Some very important data'),
              comment='A very business critical table',
              schema='s')
dialect = AthenaDialect()
print(str(dialect.ddl_compiler(dialect, CreateTable(table))))
```
Would produce the following SQL statement:
```sql
CREATE EXTERNAL TABLE s.table_name (
    c INTEGER COMMENT 'Some very important data'
)
COMMENT 'A very business critical table'
```

References:
- [Athena's `CREATE TABLE` statement documentation](https://docs.aws.amazon.com/athena/latest/ug/create-table.html)

laughingman7743:
Thanks for the pull request.

I would not accept any way to parse the result of DESCRIBE TABLE. If there is any change in the output, an error will occur. If you want to get the comments of the table, you can use the information schema.
https://docs.aws.amazon.com/athena/latest/ug/querying-glue-catalog.html

I'm not sure if the location and compression settings can be retrieved from the information schema, but Boto3 can retrieve the table metadata.
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/athena.html#Athena.Client.get_table_metadata
If we can't get this information from the information schema, I think it's best to use the API, because inside the cursor we are just sending a request to the AWS API, and it's easy to implement to get the data from this API.

laughingman7743:
Here is the branch where I tried to implement the table metadata retrieval in the fix of #244.
https://github.com/laughingman7743/PyAthena/pull/265
Please check it out.

PR #265, opened 2022-01-16
Support getting view names in SQLAlchemy (fix #244)

### LINKED ISSUE #244: View detection support
When using PyAthena with SQLAlchemy, I noticed that `get_table_names()` after inspection returns all tables _and views_ while `get_view_names()` raises a `NotImplementedError`. Is this expected? This is not critical for my use case but I'm happy to send a PR if someone can point me in the right direction.

Thanks!

laughingman7743:
TODO:
- [x] Unit tests

### DIFF
--- a/pyathena/connection.py
+++ b/pyathena/connection.py
@@ -51,7 +51,7 @@ class Connection(object):
         s3_staging_dir: Optional[str] = None,
         region_name: Optional[str] = None,
         schema_name: Optional[str] = "default",
-        catalog_name: Optional[str] = None,
+        catalog_name: Optional[str] = "awsdatacatalog",
         work_group: Optional[str] = None,
         poll_interval: float = 1,
         encryption_option: Optional[str] = None,
--- a/pyathena/model.py
+++ b/pyathena/model.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 import logging
 from datetime import datetime
-from typing import Any, Dict, Optional
+from typing import Any, Dict, List, Optional
 
 from pyathena.error import DataError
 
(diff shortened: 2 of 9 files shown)


pyathena/model.py
```diff
@@ -160,6 +160,98 @@ def work_group(self) -> Optional[str]:
...
+    @property
+    def columns(self) -> List[AthenaTableMetadataColumn]:
+        return self._columns
+
+    @property
+    def partition_keys(self) -> List[AthenaTableMetadataPartitionKey]:
+        return self._partition_keys
+
+    @property
+    def parameters(self) -> Dict[str, str]:
```
laughingman7743, line 251: It may be a good idea to make specific items, such as locaiton and comment, into properties, as shown below.
```
@property
def location(self) -> Optional[str]:
    return self._parameters.get("location", None)
```

laughingman7743: added: https://github.com/laughingman7743/PyAthena/pull/265/commits/e6bbf02572a1c5a3a355ac22e0ef475c38a21160

tests/test_sqlalchemy_athena.py
```diff
@@ -34,13 +34,6 @@
 
 
 class TestSQLAlchemyAthena(unittest.TestCase):
-    """Reference test case is following:
-
-    https://github.com/dropbox/PyHive/blob/master/pyhive/tests/sqlalchemy_test_case.py
-    https://github.com/dropbox/PyHive/blob/master/pyhive/tests/test_sqlalchemy_hive.py
-    https://github.com/dropbox/PyHive/blob/master/pyhive/tests/test_sqlalchemy_presto.py
-    """
```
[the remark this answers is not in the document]
laughingman7743: Yes, it is. Many of the implementations in this library are built with PyHive as a reference, and it should be good to mention it in the README.

pyathena/connection.py
```diff
@@ -51,7 +51,7 @@ def __init__(
         s3_staging_dir: Optional[str] = None,
         region_name: Optional[str] = None,
         schema_name: Optional[str] = "default",
-        catalog_name: Optional[str] = None,
+        catalog_name: Optional[str] = "awsdatacatalog",
```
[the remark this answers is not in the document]
laughingman7743: For the default catalog, there is no need to specify it, so Optional is fine.

tests/test_sqlalchemy_athena.py
```diff
@@ -167,6 +159,15 @@ def test_get_table_names(self, engine, conn):
             insp.get_table_names(schema=SCHEMA),
```
cansjt, line 159: With respect to the issue reported, we should also check that no view is listed:
```python
         table_names = insp.get_table_names(schema=SCHEMA)
         self.assertIn("many_rows", table_names)
         self.assertNotIn("view_one_row", table_names)
```

pyathena/sqlalchemy_athena.py
```diff
@@ -372,20 +372,34 @@ def get_schema_names(self, connection, **kw):
                 """
         return [row.schema_name for row in connection.execute(query).fetchall()]
 
-    @reflection.cache
-    def get_table_names(self, connection, schema=None, **kw):
+    def _get_tables(self, connection, schema=None, **kw):
```
cansjt, line 375: Looking into the reflection cache implementation, it really is a [basic memoization](https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/engine/reflection.py#L43) (capture a function name and its args as key, mapped to the function call result). 
So if we want to make the best (avoid as many Ahtena API calls), I think we should place the here.
```suggestion
    @reflection.cache
    def _get_tables(self, connection, schema=None, **kw):
```
Only remains in the `get_view_names()` and `get_table_names()` functions the filtering of the global list of metadata.
We could cache that also, but the trade-off between memory used and time-gain is less obvious.

laughingman7743: OK, that looks good.

cansjt (approved):
I think with this, I have most, if not all, of what I need for #263. 

Thanks.

laughingman7743:
Thanks for the review. I will merge it once all the tests have passed.

PR #264, opened 2022-01-15
Add support for table and column comments

cansjt:
As mentioned in the commit comments, beyond the support of comments, by doing this we all put in place the necessary introspection mechanism need for implementing table properties and storage formats other than parquet.

PR #269, opened 2022-01-16
Add partitions support

cansjt:
Adds support for creating partitioned tables.

Tests pass with Python 3.8 & 3.9.

### DIFF
--- a/pyathena/sqlalchemy_athena.py
+++ b/pyathena/sqlalchemy_athena.py
@@ -482,8 +482,9 @@ class AthenaDialect(DefaultDialect):
 
     def has_table(self, connection, table_name, schema=None, **kw):
         try:
-            columns = self.get_columns(connection, table_name, schema)
-            return True if columns else False
+            return (
+                self._get_table(connection, table_name, schema=None, **kw) is not None
+            )
         except NoSuchTableError:
             return False
 
--- a/pyathena/sqlalchemy_athena.py
+++ b/pyathena/sqlalchemy_athena.py
@@ -187,6 +187,29 @@ class AthenaDDLCompiler(DDLCompiler):
 
         return f"'{value}'"
 
+    def _get_columns_specs(self, create, partitions=False):
+        text = ""
+        separator = "\n"
+        for create_column in create.columns:
+            column = create_column.element
+            if column.dialect_options["awsathena"]["partition"] is not partitions:
+                continue
+            try:
+                processed = self.process(create_column)
+                if processed is not None:
+                    text += separator
+                    separator = ", \n"
+                    text += "\t" + processed
+            except exc.CompileError as ce:
+                util.raise_from_cause(
+                    exc.CompileError(
+                        util.u("(in table '{0}', column '{1}'): {2}").format(
+                            create.element.description, column.name, ce.args[0]
+                        )
+                    )
+                )
+        return text
+
     def get_column_specification(self, column, **kwargs):
         colspec = (
             self.preparer.format_column(column)
@@ -207,22 +230,7 @@ class AthenaDDLCompiler(DDLCompiler):
         text += "TABLE " + preparer.format_table(table) + " ("
 
         separator = "\n"
-        for create_column in create.columns:
-            column = create_column.element
-            try:
-                processed = self.process(create_column)
-                if processed is not None:
-                    text += separator
-                    separator = ", \n"
-                    text += "\t" + processed
-            except exc.CompileError as ce:
-                util.raise_from_cause(
-                    exc.CompileError(
-                        util.u("(in table '{0}', column '{1}'): {2}").format(
-                            table.description, column.name, ce.args[0]
-                        )
-                    )
-                )
+        text += self._get_columns_specs(create)
 
         const = self.create_table_constraints(
             table,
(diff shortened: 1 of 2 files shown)


pyathena/sqlalchemy_athena.py
```diff
@@ -398,69 +498,53 @@ def get_view_names(self, connection, schema=None, **kw):
...
+            columns_specs.append({
+                "name": column.name,
+                "type": _TYPE_MAPPINGS.get(
+                    self._get_column_type(column.type), NULLTYPE
+                ),
+                "nullable": True,  # True if row.is_nullable == "YES" else False,
+                "default": None,
+                # "default": row.column_default
+                # if not self._is_nan(row.column_default)
+                # else None,
```
cansjt, line 530: Removed `'ordinal_position'`, not part of [the specification](https://docs.sqlalchemy.org/en/14/core/reflection.html?highlight=get_columns#sqlalchemy.engine.reflection.Inspector.get_columns) (for neither of SQLAlchemy 1.3, 1.4 or 2.0). Also given that Athena does not support any of `default` or `nullable`, lets not pretend it does.

- [ ] Clean-up

```suggestion
                "nullable": True,
                "default": None,
```

pyathena/sqlalchemy_athena.py
```diff
@@ -320,7 +410,7 @@ class AthenaDialect(DefaultDialect):
     _pattern_data_catlog_exception = re.compile(
         r"(((Database|Namespace)\ (?P<schema>.+))|(Table\ (?P<table>.+)))\ not\ found\."
     )
-    _pattern_column_type = re.compile(r"^([a-zA-Z]+)($|\(.+\)$)")
+    _pattern_column_type = re.compile(r"^([a-zA-Z]+)($|[(<].+[)>]$)")
```
cansjt, line 413: The API returns the same syntax as DDL statements for composite types, e.g. `array<int>` instead of `array(int)` when using SQL & the information schema.

pyathena/sqlalchemy_athena.py
```diff
@@ -272,17 +352,21 @@ def post_create_table(self, table):
...
     "varchar": STRINGTYPE,
     "array": STRINGTYPE,
     "row": STRINGTYPE,  # StructType
+    "struct": STRINGTYPE,  # StructType
     "varbinary": BINARY,
+    "binary": BINARY,
     "map": STRINGTYPE,
     "date": DATE,
     "timestamp": TIMESTAMP,
+    "string": STRINGTYPE,
```
cansjt, line 369: Type names that differs when using the API compared to the information schema.
Should we remove the old ones ?
```suggestion
    "int": INTEGER,
    "bigint": BIGINT,
    "decimal": DECIMAL,
    "char": STRINGTYPE,
    "varchar": STRINGTYPE,
    "array": STRINGTYPE,
    "struct": STRINGTYPE,  # StructType
    "binary": BINARY,
    "map": STRINGTYPE,
    "date": DATE,
    "timestamp": TIMESTAMP,
    "string": STRINGTYPE,
```

pyathena/sqlalchemy_athena.py
```diff
@@ -211,18 +275,34 @@ def visit_create_table(self, create):
         if const:
             text += separator + "\t" + const
 
-        text += "\n)\n%s\n\n" % self.post_create_table(table)
+        text += "\n)\n%s\n\n" % self.post_create_table(create)
         return text
 
-    def post_create_table(self, table):
+    def post_create_table(self, create):
```
cansjt, line 281: Shouldn't we make this private?
```suggestion
    def _post_create_table(self, create):
```

pyathena/sqlalchemy_athena.py
```diff
@@ -398,69 +498,53 @@ def get_view_names(self, connection, schema=None, **kw):
...
+            columns_specs.append({
+                "name": column.name,
+                "type": _TYPE_MAPPINGS.get(
+                    self._get_column_type(column.type), NULLTYPE
+                ),
+                "nullable": True,  # True if row.is_nullable == "YES" else False,
+                "default": None,
+                # "default": row.column_default
+                # if not self._is_nan(row.column_default)
+                # else None,
```
cansjt, line 543: Same as above.

- [ ] Clean-up.

laughingman7743:
**Please don't include the change of getting metadata from the information schema to the API in this pull request. If you need to change it, please do so in a separate pull request. This pull request should focus on support for partitions in SQLAlchemy.**

laughingman7743:
https://github.com/laughingman7743/PyAthena/pull/264#pullrequestreview-857023832
**Please set up GitHubActions to run tests in the forked repository.**
If these pull requests continue, I will block you.

cansjt:
So I looked into setting up github actions. There are two points that make it a bit problematic for us (and I guess for anyone would want to fork the repository):
- the scheduled workflow. If it could be split from the "push" test workflow that would be nice. I know it is not very convenient, because it would mean having two almost identical workflows in separate files. Or maybe there is a smarter way to do this. I am not very familiar with Github Action. We could script something to ease the maintenance. To the least, if it were possible to significantly reduce it's frequency (_e.g._ to once a week);
- the fact that it is assumed in the tests are run by the default work group, some tests rely on that. In the company where I work all those default resources are disabled, I cannot use them. There is an `AWS_ATHENA_WORK_GROUP` environment variable we could set (to 'primary' if that's what fit you environment). And get it's value in the tests to not hard code it, as is done for the `AWS_ATHENA_S3_STAGING_DIR` variable;

Would you be open to that? Thanks.

Issue #254, opened 2021-11-25
SQLAlchemy + Pandas very slow when compared to AWS Wrangler


avibrazil:
Tested same query that returns more than 2 million lines.

AWS Wrangler takes 1m34s to return a DataFrame.
SQLAlchemy + PyAthena takes 16m37s to return a DataFrame.

[See attached notebook for proof and methods.](https://github.com/laughingman7743/PyAthena/files/7604627/PyAthena.with.SQLAlchemy.versus.AWS.Wrangler.pdf)
.

laughingman7743:
https://github.com/laughingman7743/PyAthena/issues/46
https://github.com/laughingman7743/PyAthena/tree/master/benchmarks
https://github.com/laughingman7743/PyAthena#pandascursor

avibrazil:
Thank you. I’ve seen all these links and methods before.

But they won’t clarify how to use this with SQLAlchemy. And I need to use SQLAlchemy.

So, is there any way to use PyAthena with SQLAlchemy and Pandas that would match acceptable performance?

Maybe a flag in the URL or something.

laughingman7743:
I do not have an implementation that uses a combination of SQLAlchemy and Pandas. What is the use case for that?

avibrazil:
I have a Pandas application that is independent of database. So SQLAlchemy provides a unified single interface and code doesn’t need to be changed when changing database backend, just the URL.

laughingman7743:
For now, the current implementation does not allow the use of PandasCursor and SQLAlchemy in combination.
It needs some modification, but by implementing switching the cursor to be used by the connection URL, we may possibly speed up the retrieval of query results.

avibrazil:
Are there any guides on how to use it? I’m using SQLALCHEMY query builder to query Athena through pandas read_sql_query().

Also, I’d suggest to be more explicit about this problem when using PyAthena with SQLALCHEMY and Pandas. Simply to suggest to not use SQLALCHEMY, to bypass SQLALCHEMY, is not a solution.

laughingman7743:
What do you need a guide for, a guide for PandasCursor?
It has a DB-API2.0 interface, which is the same as the default cursor. The way to fetch data is also the same.

avibrazil:
I don't know how to use it in this situation:

```python
conn_str = "awsathena+rest://:@athena.{region_name}.amazonaws.com:443/{schema_name}?work_group={work_group}&compression=snappy"

[...]

PR #271, opened 2022-01-23
Add AthenaPandas dialect (refs: #254)

laughingman7743:


cansjt:
I am really not sure about this. Dialects are meant to represent different ways to interact with the query engine. But it has nothing to do with the way query results are represented. But they allow you to make that "customizable", via dependency injection.

If you do this, your coupling things that should remain decoupled, so as to leave as many options as possible to users. How would you represent using the JDBC driver with Pandas? (would it exists) Or the JDBC or REST driver with any other kind of cursor? Or what if people want the asynchronous version of the cursor?

Passing `cursor_class` on the connection URL should be enough `awsathena+rest://.../database?cursor_class=pyathena.pandas.cursor.PandasCursor`, if that makes your life easier (to not have to pass it to `create_engine()` within code, but add that to some configuration).

All you need to make this work, is to add something this:
```diff
diff --git a/pyathena/sqlalchemy_athena.py b/pyathena/sqlalchemy_athena.py
index 1a7fa1a..c378e3b 100644
--- a/pyathena/sqlalchemy_athena.py
+++ b/pyathena/sqlalchemy_athena.py
@@ -3,6 +3,7 @@
 import numbers
 import re
 from distutils.util import strtobool
+import importlib
 
 import botocore
 from sqlalchemy import exc, schema, util
@@ -409,6 +410,12 @@ def create_connect_args(self, url):
             opts.update(
                 {"kill_on_interrupt": bool(strtobool(url.query["kill_on_interrupt"]))}
             )
+        if "cursor_class" in opts:
+            module_name, class_name = opts['cursor_class'].rsplit('.', maxsplit=1)
+            # Raises ImportError if the module does not exist
+            module = importlib.import_module(module_name)
+            # Raises AttributeError in case the class doesn't exist
+            opts['cursor_class'] = getattr(module, class_name)
         return [[], opts]
 
     @reflection.cache
```
And you're free to handle "shorthands" if you'd like. E.g. mapping a string like `'asyncpandas'` to the actual class. That's even better than the above, in that it hides some implementation details (the exact class name) from the user, leaving you free to refactor your code at will, would the need arise.

The above first is better, though, if you intend to let people implement their own cursor class. But combining both approaches is not that tough.
```diff
diff --git a/pyathena/sqlalchemy_athena.py b/pyathena/sqlalchemy_athena.py
index 1a7fa1a..6a85342 100644
--- a/pyathena/sqlalchemy_athena.py
+++ b/pyathena/sqlalchemy_athena.py
@@ -3,6 +3,7 @@
 import numbers
 import re
 from distutils.util import strtobool
+import importlib
 
 import botocore
 from sqlalchemy import exc, schema, util
@@ -29,6 +30,13 @@
 )
 
 import pyathena
+from pyathena.pandas import async_cursor, cursor
+
+
+KNOWN_CURSOR_IMPL = {
+    "pandas": cursor.PandasCursor,
+    "asyncpandas": async_cursor.AsyncPandasCursor,
+}
 
 
 class UniversalSet(object):
@@ -409,6 +417,18 @@ def create_connect_args(self, url):
             opts.update(
                 {"kill_on_interrupt": bool(strtobool(url.query["kill_on_interrupt"]))}
             )
+        if "cursor_class" in opts:
+            cursor_class = KNOWN_CURSOR_IMPL.get(opts["cursor_class"])
+            if cursor_class:
+                opts["cursor_class"] = cursor_class
+            else:
+                module_name, class_name = opts.pop("cursor_class").rsplit(
+                    ".", maxsplit=1
+                )
+                # Raises ImportError if the module does not exist
+                module = importlib.import_module(module_name)
+                # Raises AttributeError in case the class does not exist
+                opts["cursor_class"] = getattr(module, class_name)
         return [[], opts]
 
     @reflection.cache
```

laughingman7743:
The driver specification is not necessarily in the form of a protocol name, such as pymysql, mariadbconnector, or mysqldb in the dialect included with SQLAlchemy. (The name of the DB-API is specified.)

PandasCursor is currently being implemented quite aggressively to comply with the DB-API. We believe that a special implementation may be needed in the future to support this cursor.

Of course, the implementation you suggest is possible, but I think it would be better to switch explicitly in the connection string schema, so that we can implement our own without worrying too much about matching the internal implementation. In addition, it should be easier for users to just switch the schema of the connection string.

If users want to create their own dialect for awsathena, they can easily extend it by creating a derived class of AthenaDialect with PyAthena as a dependency and adding their own driver name.
```
class AthenaXXXDialect(AthenaDialect):
    driver = "xxx"

    # A custom implementation that overrides the AthenaDialect implementation.


class AthenaDialect_foobar(AthenaDialect):
    driver = "foobar"

    # A custom implementation that overrides the AthenaDialect implementation.
```
There should be no need for tricky implementations such as import_module.

PR #281, opened 2022-02-20
Remove table constraints from CREATE TABLE statements (fix #279)

### LINKED ISSUE #279: Creating tables with SQLAlchemy models supported?
The first issue I encountered:

sqlalchemy.exc.CompileError: You need to specify the storage location for the table using the `awsathena_location` dialect keyword argument

I was able to address this with something like:

Base.metadata.tables['my_table'].dialect_options['awsathena']['location'] = location

There is probably a better solution, but this is all I could come up with so far.

Once I fixed this, I encountered this:

sqlalchemy.exc.DatabaseError: (pyathena.error.DatabaseError) An error occurred (InvalidRequestException) when calling the StartQueryExecution operation: line 1:8: mismatched input 'EXTERNAL'. Expecting: 'OR', 'SCHEMA', 'TABLE', 'VIEW'

I debugged this on the Athena console and determined the issue is the SQL generated has primary key constraints which are not supported by Athena. Removing the primary key constraint in SQLAlchemy gives:

sqlalchemy.exc.ArgumentError: Mapper mapped class Institution->institution could not assemble any primary key columns for mapped table 'my_table'

Am I in unsupported territory or am I doing something wrong?

laughingman7743:
* Remove table constraints from CREATE TABLE statements
* Change to use INT as the column type for Integer type in DDL
* Support for the If_not_exists option


laughingman7743:
TODO:
- [ ] Unit tests

PR #307, opened 2022-05-06
Implement unload options in PandasCursor (fix #242)

### LINKED ISSUE #242: PandasCursor: Add support for UNLOAD and resulting file types
Now that [Athena supports UNLOAD in the queries](https://docs.aws.amazon.com/athena/latest/ug/unload.html), it means a user can have their result in Apache Parquet, ORC, Apache Avro, or JSON file. PandasCursor currently supports only csv files.

laughingman7743:


laughingman7743:
https://github.com/laughingman7743/PyAthena/pull/325

PR #343, opened 2022-07-10
Change the Parquet library used by Pandas to allow choice of PyArrow or fastparquet (fix #341)

### LINKED ISSUE #341: PyArrow Requirement Exceeds Lambda Unzipped Max
The `PandasCursor` is no longer usable in **AWS Lambdas** because of the requirement to use [PyArrow](https://pypi.org/project/pyarrow/). The library is quite large, and it exceeds [AWS Unzipped file size limit](https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-limits.html). It is prevent deploys at the moment. 

PyArrow is quite large, and other libraries have run into similar issues: https://github.com/snowflakedb/snowflake-connector-python/issues/213

I wonder if could be made an optional dependency for the `PandasCursor`? The library could use the older logic as fallback for when it isn't present.

image

laughingman7743:


laughingman7743:
TODO:
- [ ] Unit tests

PR #370, opened 2022-09-23
Implement s3 filesystem

laughingman7743:


laughingman7743:
Performance is poor and needs to be improved. Methods such as find and exists also need to be implemented specifically for S3.
- https://github.com/fsspec/s3fs/blob/2022.11.0/s3fs/core.py#L737
- https://github.com/fsspec/s3fs/blob/2022.11.0/s3fs/core.py#L933
- https://github.com/fsspec/s3fs/blob/2022.11.0/s3fs/core.py#L1178

PR #383, opened 2022-11-25
Adding user-agent-extra parameter to improve Log Analysis

armaseg:
User-agent is key parameter to improve Log Analysis. Previously, PyAthena users and tools developed using PyAthena were not able to differentiate boto3 calls from their calls. 

Introducing the user-agent-extra parameter to the PyAthena/boto3 connections allow users to identify HTTP calls from PyAthena and tools developed using PyAthena

Example of new user-agent:
`'User-Agent': b'Boto3/1.24.59 Python/3.7.10 Linux/4.14.296-222.539.amzn2.x86_64 Botocore/1.27.59 PyAthena/2.15.0 dbt-athena-community/1.3.2'`

### DIFF
--- a/pyathena/connection.py
+++ b/pyathena/connection.py
@@ -141,7 +141,9 @@ class Connection:
 
         self.config: Optional[config.Config] = config
         self.config.user_agent_extra = (
-            "PyAthena/" + pyathena.__version__ + " " + str(config.user_agent_extra or "")
+            "PyAthena/"
+            + pyathena.__version__
+            + (" " + config.user_agent_extra if config.user_agent_extra else "")
         )
         self._client = self._session.client(
             "athena", region_name=self.region_name, config=self.config, **self._client_kwargs
--- a/pyathena/connection.py
+++ b/pyathena/connection.py
@@ -5,7 +5,9 @@ import time
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Type
 
 from boto3.session import Session
+from botocore import config
 
+import pyathena
 from pyathena.common import BaseCursor
 from pyathena.converter import Converter
 from pyathena.cursor import Cursor
@@ -35,7 +37,6 @@ class Connection:
         "aws_access_key_id",
         "aws_secret_access_key",
         "aws_session_token",
-        "config",
         "api_version",
         "use_ssl",
         "verify",
@@ -65,6 +66,7 @@ class Connection:
         cursor_kwargs: Optional[Dict[str, Any]] = None,
         kill_on_interrupt: bool = True,
         session: Optional[Session] = None,
+        config: Optional[config.Config] = config.Config(),
         **kwargs,
     ) -> None:
         self._kwargs = {
@@ -136,8 +138,11 @@ class Connection:
                 profile_name=self.profile_name,
                 **self._session_kwargs,
             )
+
+        self.config: Optional[config.Config] = config
+        self.config.user_agent_extra = "PyAthena/" + pyathena.__version__ + " " + str(config.user_agent_extra or '')
         self._client = self._session.client(
-            "athena", region_name=self.region_name, **self._client_kwargs
+            "athena", region_name=self.region_name, config=self.config, **self._client_kwargs
         )
         self._converter = converter
         self._formatter = formatter if formatter else DefaultParameterFormatter()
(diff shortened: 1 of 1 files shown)


laughingman7743:
The current implementation of the connector passes the specific kwargs specified in the arguments to the boto3 arguments. config values are also passed, so it is easy to change the value of user_agent_extra as follows.
```
import logging
from botocore import config
from pyathena import connect

logging.basicConfig(level=logging.DEBUG)

config = config.Config(user_agent_extra="foobar")
cursor = connect(
    s3_staging_dir="s3://laughingman7743-athena/path/to/",
    region_name="us-west-2",
    config=config,
).cursor()

cursor.execute("SELECT * FROM one_row")
print(cursor.description)
print(cursor.fetchall())
```
I don't want to implement something that changes to a specific user agent on the library side. The library has already been implemented in such a way that the user can set it as needed, so please set it on the user side.

pyathena/connection.py
```diff
@@ -136,8 +138,13 @@ def __init__(
                 profile_name=self.profile_name,
                 **self._session_kwargs,
             )
+
+        self.config: Optional[config.Config] = config
+        self.config.user_agent_extra = (
+            "PyAthena/" + pyathena.__version__ + " " + str(config.user_agent_extra or "")
```
iconara, line 144: This looks like the user agent would end in a space when there was no extra info, or is there some stripping happening later that was already there from before?

armaseg: I totally missed that space: Nice spotted! 
I just fixed it.
Thanks for the feedback!

armaseg:
Hello @laughingman7743,
Sorry about the misunderstanding. 
The purpose is to be able to identify calls coming from PyAthena and separate them from calls coming from using boto directly. And to be able to identify the version of PyAthena that is being used. 
Additionally, I allow users to customize using suffix in the user-agent to identify their application (ex: using the same boto standard)

This would help organizations in multiple ways. For example, Chief Information Security Officer can validate that all applications developed in their organization are using approved versions of boto3/PyAthena as they are currently doing with boto3, without depending from their developers . Additionally, their applications can add their versions at the end of the user agent.

laughingman7743:
Okay, I understand about the use case, I want you to use f-string, so please fix it.

pyathena/connection.py
```diff
@@ -136,8 +138,15 @@ def __init__(
                 profile_name=self.profile_name,
                 **self._session_kwargs,
             )
+
+        self.config: Optional[config.Config] = config
+        self.config.user_agent_extra = (
+            "PyAthena/"
+            + pyathena.__version__
+            + (" " + config.user_agent_extra if config.user_agent_extra else "")
+        )
```
laughingman7743, line 147:
```suggestion
        self.config.user_agent_extra = (
            f"PyAthena/{pyathena.__version__}"
            f"{' ' + config.user_agent_extra if config.user_agent_extra else ''}"
        )
```

pyathena/connection.py
```diff
@@ -136,8 +138,15 @@ def __init__(
                 profile_name=self.profile_name,
                 **self._session_kwargs,
             )
+
+        self.config: Optional[config.Config] = config
```
laughingman7743, line 142:
```suggestion
        self.config: Optional[config.Config] = config if config else config.Config()
```
It may be safer to take None as an argument and initialize it in the None case to handle config more safely. (In consideration of the case where None is specified in the argument config.)

pyathena/connection.py
```diff
@@ -65,6 +66,7 @@ def __init__(
         cursor_kwargs: Optional[Dict[str, Any]] = None,
         kill_on_interrupt: bool = True,
         session: Optional[Session] = None,
+        config: Optional[config.Config] = config.Config(),
```
laughingman7743, line 69:
```suggestion
        config: Optional[config.Config] = None,
```

pyathena/connection.py
armaseg: I changed to use f-string but I'm using the self.config variable in the ternary operator. Otherwise, if parameter is None I would get:
`AttributeError: 'NoneType' object has no attribute 'user_agent_extra'`

pyathena/connection.py
armaseg: I added the changes, but I also simplified the import
`from botocore.config import Config`

pyathena/connection.py
armaseg: I added the changes, but I also simplified the import
from botocore.config import Config

pyathena/connection.py
laughingman7743: Thanks for the update. The AttributeError should no longer occur with the following changes.
https://github.com/laughingman7743/PyAthena/pull/383#discussion_r1032624262

laughingman7743:
It will probably pass the test. I will merge it.
Thanks for your contribution, the next version v2.16.0 will be released soon.

armaseg:
Thank you @laughingman7743 😄

Issue #385, opened 2022-11-29
STS requires self.config missing attribute


maicon-berndt:
By creating a connection using ARN Role but suppressing/ignoring the optional `config` parameter PyAthena is throwing the following error message:
> AttributeError: 'Connection' object has no attribute 'config'

It happens in the line... https://github.com/laughingman7743/PyAthena/blob/15245efb8d0caaf9ccb71bc6a6ce653a090c1f76/pyathena/connection.py#L171

...once `self.config` attribute is defined with a `Config` object here... https://github.com/laughingman7743/PyAthena/blob/15245efb8d0caaf9ccb71bc6a6ce653a090c1f76/pyathena/connection.py#L142

...only after the `_assume_role` method is called: 
https://github.com/laughingman7743/PyAthena/blob/15245efb8d0caaf9ccb71bc6a6ce653a090c1f76/pyathena/connection.py#L104

I assume that the `self.config: Optional[Config] = config if config else Config()` instruction should be moved to lines above the `_assume_role` call, or the `config=self.config` parameter should be also optional by creating the `sts` client.

laughingman7743:
https://github.com/laughingman7743/PyAthena/releases/tag/v2.16.1
v2.16.1 has been released. 🎉 Please check it out.

PR #386, opened 2022-11-29
STS requires self.config missing attribute Fix

maicon-berndt:
Issue: https://github.com/laughingman7743/PyAthena/issues/385

### DIFF
--- a/pyathena/connection.py
+++ b/pyathena/connection.py
@@ -92,6 +92,7 @@ class Connection:
         self.encryption_option = encryption_option
         self.kms_key = kms_key
         self.profile_name = profile_name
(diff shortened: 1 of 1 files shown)


pyathena/connection.py
```diff
@@ -139,7 +140,6 @@ def __init__(
                 **self._session_kwargs,
             )
 
-        self.config: Optional[Config] = config if config else Config()
         self.config.user_agent_extra = (
```
maicon-berndt, line 143: I don't think these `config.user_agent_extra` lines are relevant to create the `sts` client though.
If they are relevant, I can move them also to the lines above.

laughingman7743:
@maicon-berndt Thanks!

PR #403, opened 2023-02-04
Removing filter that required tables to be explicitly external

bradleyhurley:
I don't have easy access to an AWS account for running the tests, but Im happy to make changes or update test if you can provide me with feedback.

PR #410, opened 2023-02-13
Add EXTERNAL table type supported by Athena federation to the filter

mai-nguyen-coinhako:
In Athena federation query, there is a new Table Type **EXTERNAL**. The fix from this PR https://github.com/laughingman7743/PyAthena/pull/403 in PyAthena only supports **EXTERNAL_TABLE** and **MANAGED_TABLE**, which is created via Spark when Glue is enabled as hivemeta store, though EXTERNAL tables are fully queryable in Athena.


laughingman7743:
Thank you, v2.23.0 has been released.
https://pypi.org/project/pyathena/2.23.0/

mai-nguyen-coinhako:
Thank you

PR #421, opened 2023-04-09
Add clobber=True when registering s3 impls

gallushi:
`fsspec` recently [disabled](https://github.com/fsspec/filesystem_spec/pull/1237) clobbering by default.

This causes an issue when importing `pyathena.pandas` since it tries to register an impl for s3, which already has a known impl in fsspec, so we must now explicitly pass `clobber=True` (the default behavior prior to the change).

### DIFF
--- a/pyathena/pandas/__init__.py
+++ b/pyathena/pandas/__init__.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
 import fsspec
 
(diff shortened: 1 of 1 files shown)


romeran:
+1

laughingman7743:
Thanks!

laughingman7743:
https://pypi.org/project/pyathena/2.24.0/
Just released v2.24.0. If necessary, I can release a patch version with this change applied to versions v2.18 through v2.23.

Issue #445, opened 2023-05-23
PyAthena 3.0.x doesn't work with SQLAlchemy 1.4.36


mdeshmu:
Hi,

Apache Superset uses [SQLAlchemy 1.4.36](https://github.com/apache/superset/blob/master/requirements/base.txt#L279) which when used with recently released PyAthena versions 3.0.x throws below error:

`NameError: name 'Union' is not defined`

Here is the error stack trace:
```
Traceback (most recent call last):
  File "/usr/local/bin/superset", line 33, in <module>
    sys.exit(load_entry_point('apache-superset', 'console_scripts', 'superset')())
  File "/usr/local/lib/python3.8/site-packages/click/core.py", line 1128, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/flask/cli.py", line 567, in main
    return super().main(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/click/core.py", line 1053, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.8/site-packages/click/core.py", line 1659, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.8/site-packages/click/core.py", line 1659, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.8/site-packages/click/core.py", line 1395, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.8/site-packages/click/core.py", line 754, in invoke
    return __callback(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/click/decorators.py", line 26, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/flask/cli.py", line 407, in decorator
    return __ctx.invoke(f, *args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/click/core.py", line 754, in invoke
    return __callback(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/flask_migrate/cli.py", line 149, in upgrade
    _upgrade(directory, revision, sql, tag, x_arg)
  File "/usr/local/lib/python3.8/site-packages/flask_migrate/__init__.py", line 98, in wrapped
    f(*args, **kwargs)
[...]

PR #446, opened 2023-05-24
Delay type annotation eval for compatibility with SQLAlchemy 1.4

xuganyu96:
This PR aims to address #445.

Due to the strange way SQLAlchemy wraps around PyAthena classes and functions, type hints are eagerly evaluated, which result in errors mentioned in the issue above. Adding `from __future__ import annotations` delays the evaluation of those type hints, which should not impact usage with IDE's and LSP's, but should allow newer versions of PyAthena to play nicely with SQLAlchemy 1.4.

Since `from __future__ import annotations` does not affect code evaluation and has already been use a few times elsewhere in the code base, this PR simply provides a blanket application of this state everywhere type hints are used. Please advise if we need to be more specific with applying this statement.

Thank you very much for considering this contribution.

### DIFF
--- a/pyathena/__init__.py
+++ b/pyathena/__init__.py
@@ -1,4 +1,6 @@
 # -*- coding: utf-8 -*-
+from __future__ import annotations
+
 import datetime
 from typing import TYPE_CHECKING, FrozenSet, Type
 
--- a/pyathena/arrow/async_cursor.py
+++ b/pyathena/arrow/async_cursor.py
@@ -1,4 +1,6 @@
 # -*- coding: utf-8 -*-
+from __future__ import annotations
+
 import logging
 from concurrent.futures import Future
 from multiprocessing import cpu_count
--- a/pyathena/arrow/converter.py
+++ b/pyathena/arrow/converter.py
@@ -1,4 +1,6 @@
 # -*- coding: utf-8 -*-
+from __future__ import annotations
+
 import logging
 from builtins import isinstance
 from copy import deepcopy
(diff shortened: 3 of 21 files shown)


laughingman7743:
https://github.com/laughingman7743/PyAthena/blob/master/pyathena/sqlalchemy/base.py#L4-L18
By using `from __future__ import annotations`, Union can be replaced with a piped format like `Foo | Bar`, but that change is not necessary?

xuganyu96:
> https://github.com/laughingman7743/PyAthena/blob/master/pyathena/sqlalchemy/base.py#L4-L18 By using `from __future__ import annotations`, Union can be replaced with a piped format like `Foo | Bar`, but that change is not necessary?

Thank you for the suggestion.

Union type using `|` is introduced in [PEP 604](https://peps.python.org/pep-0604/) and effective in Python 3.10. Using `|` should be exactly the same as using `Union`. There are also other type hint changes in newer (Python 3.7+) versions of Python, such as using `tuple` and `list` instead of `Tuple` and `List`, and `Optional` being equivalent to `T | None`. They are all valid and probably more concise type annotations than their legacy counterparts.

However, that is not the focus on this PR. The aim of this PR is to use the `__future__` import to resolve compatibility issues with SQL Alchemy 1.4. I'd be happy to raise a separate issue for revamping the type hints in this code base, but I'd also like to maintain the scope of this pull request.

laughingman7743:
OK, then I will merge this pull request.

Issue #458, opened 2023-07-03
Sqlalchemy Create ICEBERG table wrong Partition clause


dominik-reiner:
Hi,
I noticed that when you define a ICEBERG table in Sqlalchemy like this: ...

```
session_table = Table(
    "session"
    , metadata_obj
    , Column('session_id', String)
    , Column('event_date', Date, awsathena_partition=True)
    , awsathena_tblproperties={
        "table_type": "ICEBERG"
    }
)
```
... the PARTITION clause in the DDL SQL statement is generated incorrectly:
```
CREATE TABLE test.session (
	session_id STRING,
)
PARTITIONED BY (
	event_date DATE
)
LOCATION 's3://XXX/test/session/'
TBLPROPERTIES (
	'table_type' = 'ICEBERG'
)
```

This is due to the syntax for ICEBERG tables PARTITION clause being different in that the column has to appear first in the COLUMN clause and then just the column name in the PARTITION clause.
See here:
https://docs.aws.amazon.com/athena/latest/ug/querying-iceberg-creating-tables.html#querying-iceberg-partitioning

The Athena UI in AWS also reports:
> Invalid PARTITIONED BY clause in Iceberg create table statement

laughingman7743:
https://github.com/laughingman7743/PyAthena/blob/master/pyathena/sqlalchemy/base.py#L890-L896
Thank you, it might be necessary to add an item to construct_arguments to allow specifying a partition transform function in order to support partitions in the ICEBERG table.

dominik-reiner:
I can try to submit a PR.

laughingman7743:
I just released v3.0.5.
https://pypi.org/project/pyathena/3.0.5/
https://github.com/laughingman7743/PyAthena/releases/tag/v3.0.5

PR #459, opened 2023-07-04
Implemented ICEBERG partition & partition transform functions

dominik-reiner:
Solves #458 

I adapted the handling of the column dialect option 'partition' in _prepared_columns() for ICEBERG tables.
I also added three column dialect options. 
- "partition_transform" - pass partition transform function
- "partition_transform_bucket_count" - pass bucket count for partition transform function 'bucket'
- "partition_transform_truncate_length" - pass truncate length for partition transform function 'truncate'

All partition functions are defined in model.py under AthenaPartitionTransform.

I did not add handlers for the connect options for those column dialect options.
Also did not update the README.

PR #460, opened 2023-07-05
Implemented ICEBERG partition & partition transform functions

laughingman7743:
Solves #458 

I adapted the handling of the column dialect option 'partition' in _prepared_columns() for ICEBERG tables.
I also added three column dialect options. 
- "partition_transform" - pass partition transform function
- "partition_transform_bucket_count" - pass bucket count for partition transform function 'bucket'
- "partition_transform_truncate_length" - pass truncate length for partition transform function 'truncate'

All partition functions are defined in model.py under AthenaPartitionTransform.

I did not add handlers for the connect options for those column dialect options.
Also did not update the README.

### DIFF
--- a/pyathena/model.py
+++ b/pyathena/model.py
@@ -487,3 +487,23 @@ class AthenaCompression:
             AthenaCompression.COMPRESSION_ZLIB,
             AthenaCompression.COMPRESSION_ZSTD,
         ]
+
+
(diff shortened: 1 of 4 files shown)


laughingman7743:
https://github.com/laughingman7743/PyAthena/pull/459#issuecomment-1621585850

PR #468, opened 2023-08-23
implement `get_view_definition()` for AthenaDialect

bossenti:
This PR implements the `get_view_definition` method for the `AthenaDialect` class.
The `get_view_definition` method is already provided by SQLAlchemy's [`Default` class](https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/engine/interfaces.py#L1544) in an abstract manner.

I hope this PR works out well for you.
Please let me know if there is anything to improve or adapt.


laughingman7743:
Thanks for the contribution. I will check it.

pyathena/sqlalchemy/base.py
```diff
@@ -1103,6 +1103,15 @@ def has_table(
         except NoSuchTableError:
             return False
 
+    @reflection.cache
+    def get_view_definition(self, connection, view_name: str, schema: str, **kw):
+        query = f"""SHOW CREATE VIEW "{schema}"."{view_name}";"""
+        res = connection.scalars(text(query))
+        if res is None:
```
laughingman7743, line 1110: If the View does not exist, the query result will not be None, but an OperationalError will be generated. I do not think that all cases where an OperationalError occurs are cases where the View does not exist, but if an OperationalError occurs, I think it is fine to generate a NoSuchTableError.

tests/pyathena/sqlalchemy/test_base.py
```diff
@@ -1796,3 +1796,23 @@ def test_insert_from_select_cte_follows_insert_one(self, engine):
...
+            "FROM",
+            f"  {ENV.schema}.one_row",
+        ]
+                )
+
+    def test_get_view_definition_missing_view(self, engine):
+        engine, conn = engine
+        insp = sqlalchemy.inspect(engine)
+        pytest.raises(
+            OperationalError,
```
laughingman7743, line 1816: I would like to see NoSuchTableError checked.

tests/pyathena/sqlalchemy/test_base.py
```diff
@@ -1796,3 +1796,23 @@ def test_insert_from_select_cte_follows_insert_one(self, engine):
...
+            lambda: insp.get_view_definition(schema=ENV.schema, view_name="test-view")
```
laughingman7743, line 1817: NIT:
```suggestion
            lambda: insp.get_view_definition(schema=ENV.schema, view_name="test_view")
```

laughingman7743: Applying `make fmt` will add a trailing comma. Please format your code with `make fmt`.

bossenti:
Thanks for reviewing @laughingman7743 
I will go through your comments and address them

pyathena/sqlalchemy/base.py
bossenti: I have followed the implementation of `sqlalchemy` here to stay consistent. See these two examples for this:
* [Postgres](https://github.com/sqlalchemy/sqlalchemy/blob/7cf836241aa72494b42b645a3ddec04b0a6e6061/lib/sqlalchemy/dialects/postgresql/base.py#L3438)
* [Oracle](https://github.com/sqlalchemy/sqlalchemy/blob/7cf836241aa72494b42b645a3ddec04b0a6e6061/lib/sqlalchemy/dialects/oracle/base.py#L3136)

laughingman7743: Does the same OperationalError occur in Oracle or Postgres when the View does not exist? 🤔

tests/pyathena/sqlalchemy/test_base.py
bossenti: sure 🙈

tests/pyathena/sqlalchemy/test_base.py
bossenti: Done

pyathena/sqlalchemy/base.py
bossenti: Ah, now I see what you mean 😅. 
You are right, I should probably get rid of `if res is None` and replace it with a `try: ... except OperationalError` statement.

tests/pyathena/sqlalchemy/test_base.py
```diff
@@ -11,7 +11,7 @@
 import pytest
 import sqlalchemy
 from sqlalchemy import types
-from sqlalchemy.exc import NoSuchTableError
+from sqlalchemy.exc import NoSuchTableError, OperationalError
```
laughingman7743, line 14:
```suggestion
from sqlalchemy.exc import NoSuchTableError
```

pyathena/sqlalchemy/base.py
```diff
@@ -1103,6 +1103,17 @@ def has_table(
         except NoSuchTableError:
             return False
 
+    @reflection.cache
+    def get_view_definition(self, connection, view_name: str, schema: str, **kw):
+        query = f"""SHOW CREATE VIEW "{schema}"."{view_name}";"""
+
+        try:
+            res = connection.scalars(text(query))
+        except OperationalError as exc:
```
laughingman7743, line 1112: OperationalError import is required.

laughingman7743: You can easily run lint with `make chk`.

pyathena/sqlalchemy/base.py
```diff
@@ -19,7 +19,7 @@
 )
 
 import botocore
-from sqlalchemy import exc, schema, types, util
+from sqlalchemy import exc, schema, text, types, util
 from sqlalchemy.engine import Engine, reflection
 from sqlalchemy.engine.default import DefaultDialect
 from sqlalchemy.exc import NoSuchTableError
```
laughingman7743, line 25:
```suggestion
from sqlalchemy.exc import NoSuchTableError, OperationalError
```

pyathena/sqlalchemy/base.py
laughingman7743:
```suggestion
        except OperationalError as e:
```
The error occurs because the Exception alias is a duplicate of import exc.

laughingman7743:
```
pyathena/sqlalchemy/base.py:1107: error: Signature of "get_view_definition" incompatible with supertype "Dialect"  [override]
pyathena/sqlalchemy/base.py:1107: note:      Superclass:
pyathena/sqlalchemy/base.py:1107: note:          def get_view_definition(self, connection: Connection, view_name: str, schema: Optional[str] = ..., **kw: Any) -> str
pyathena/sqlalchemy/base.py:1107: note:      Subclass:
pyathena/sqlalchemy/base.py:1107: note:          def get_view_definition(self, connection: Any, view_name: str, schema: str, **kw: Any) -> Any
pyathena/sqlalchemy/base.py:1113: error: "OperationalError" has no attribute "NoSuchTableError"  [attr-defined]
Found 2 errors in 1 file (checked 36 source files)
```

bossenti: > You can easily run lint with make chk.

Sorry for forgetting about it 

> Superclass:
pyathena/sqlalchemy/base.py:1107: note:          def get_view_definition(self, connection: Connection, view_name: str, schema: Optional[str] = ..., **kw: Any) -> str

Interesting. So it seems that the abstract implementation expects a string return type.
However, the two examples I shared from sqlalchemy don't seem to do this (at least to my limited understanding).

So I would stick to the superclass definition and convert the result to a string.
What do you think?

laughingman7743: Yes, that would be fine.

pyathena/sqlalchemy/base.py
```diff
@@ -1102,6 +1102,26 @@ def has_table(
...
+        try:
+            res = connection.execute(text(query))
+        except exc.OperationalError as e:
+            raise exc.NoSuchTableError(f"{schema}.{view_name}" if schema else view_name) from e
+        else:
+            statement_rows = [r for r in res]
+
+            view_statement = "\n".join([row[0] for row in statement_rows])
+
+            return view_statement
```
laughingman7743, line 1123:
```suggestion
        raw_connection = self._raw_connection(connection)
        schema = schema if schema else raw_connection.schema_name  # type: ignore
        query = f"""SHOW CREATE VIEW "{schema}"."{view_name}";"""
        try:
            res = connection.scalars(text(query))
        except exc.OperationalError as e:
            raise exc.NoSuchTableError(f"{schema}.{view_name}") from e
        else:
            return "\n".join([r for r in res])
```
I think converting Athena results to strings could be more easily implemented. The schema is always specified, so you should be able to get it from raw_connection if it is not specified like the other methods.

I am not sure if it is better to concatenate the results using newlines and return them, but it seems like a good idea for now.

tests/pyathena/sqlalchemy/test_base.py
```diff
@@ -1796,3 +1796,20 @@ def test_insert_from_select_cte_follows_insert_one(self, engine):
...
         assert actual == [(2, "bar")]
+
+    def test_get_view_definition(self, engine):
+        engine, conn = engine
+        insp = sqlalchemy.inspect(engine)
+        actual = insp.get_view_definition(schema=ENV.schema, view_name="v_one_row")
+        assert (
+            actual == f"CREATE VIEW {ENV.schema}.v_one_row AS\nSELECT number_of_rows\n"
+            f"FROM\n  {ENV.schema}.one_row"
+        )
```
laughingman7743, line 1807:
```suggestion
        assert (
            actual
            == textwrap.dedent(
                f"""
            CREATE VIEW {ENV.schema}.v_one_row AS
            SELECT number_of_rows
            FROM
              {ENV.schema}.one_row
            """
            ).strip()
        )
```

laughingman7743:
https://pypi.org/project/PyAthena/3.0.7/
https://github.com/laughingman7743/PyAthena/releases/tag/v3.0.7
v3.0.7 has been released. 🎉

PR #488, opened 2023-10-29
Support for Python 3.12 (fix #482)

### LINKED ISSUE #482: Support for Python 3.12
https://www.python.org/downloads/release/python-3120/

laughingman7743:


pyproject.toml
edgarrmondragon, line 1: Numpy is a bit problematic to manage for multiple Python versions with Poetry, so perhaps you wanna do something like

```toml
numpy = [
    { version = ">=1.22,<1.25", python = ">=3.8,<3.9", optional = true },
    { version = ">=1.22", python = ">=3.9", optional = true },
]
```
?

laughingman7743: Thank you, the test was successful with Python 3.12, but for some reason it takes about 10 minutes to install the dependency library.
https://github.com/laughingman7743/PyAthena/actions/runs/7308528447/job/19915228707?pr=488

laughingman7743:
```
Configuration file exists at /Users/foobar/Library/Preferences/pypoetry, reusing this directory.

Consider moving TOML configuration files to /Users/foobar/Library/Application Support/pypoetry, as support for the legacy directory will be removed in an upcoming release.
Installing dependencies from lock file

Package operations: 16 installs, 0 updates, 0 removals

  • Installing cramjam (2.7.0): Failed

  ChefBuildError

  Backend subprocess exited when trying to invoke build_wheel

  Running `maturin pep517 build-wheel -i /var/folders/y7/7k9v23tx57j82b5hvb09nggw0000gp/T/tmpqvq2yk7j/.venv/bin/python --compatibility off`
  💥 maturin failed
    Caused by: Cargo metadata failed. Do you have cargo in your PATH?
    Caused by: No such file or directory (os error 2)
  Error: command ['maturin', 'pep517', 'build-wheel', '-i', '/var/folders/y7/7k9v23tx57j82b5hvb09nggw0000gp/T/tmpqvq2yk7j/.venv/bin/python', '--compatibility', 'off'] returned non-zero exit status 1


  at ~/.pyenv/versions/3.12.1/lib/python3.12/site-packages/poetry/installation/chef.py:164 in _prepare
      160│
      161│                 error = ChefBuildError("\n\n".join(message_parts))
      162│
      163│             if error is not None:
    → 164│                 raise error from None
      165│
      166│             return path
      167│
      168│     def _prepare_sdist(self, archive: Path, destination: Path | None = None) -> Path:

Note: This error originates from the build backend, and is likely not a problem with poetry but with cramjam (2.7.0) not supporting PEP 517 builds. You can verify this by running 'pip wheel --no-cache-dir --use-pep517 "cramjam (==2.7.0)"'.
```
On macOS, installation of cramjam fails.

laughingman7743:
https://github.com/laughingman7743/PyAthena/pull/488

PR #497, opened 2024-01-05
Implement SparkCursor to support Spark calculations (fix #493)

### LINKED ISSUE #493: Add support for Spark calculations
# Description

Add support to run spark calculations using any cursor

# Related docs

1. Implementation in awswrangler - https://github.com/aws/aws-sdk-pandas/blob/main/awswrangler/athena/_spark.py
2. Official AWS docs - https://docs.aws.amazon.com/athena/latest/ug/notebooks-spark.html
3. MR in dbt-athena-community - https://github.com/dbt-athena/dbt-athena/pull/248

# Comments

I am currently working on adding support to run python models using the dbt-athena-community adapter and it would be much easier to accomplish if the pyathena library supports this first. I don't think mock_athena supports these yet so testing it actually much more difficult than I thought.

laughingman7743:


pyathena/calc_cursor.py
```diff
@@ -0,0 +1,200 @@
...
+    AthenaCalculationExecution,
+    AthenaQueryExecution,
+    AthenaSession,
+)
+from pyathena.util import retry_api_call
+
+_logger = logging.getLogger(__name__)  # type: ignore
+
+
+class CalcCursor(BaseCursor):
```
Avinash-1394, line 21: suggestion (non-blocking): I think it might be better to name it as SparkCursor to be more explicit but I will leave that up to you

pyathena/calc_cursor.py
```diff
@@ -0,0 +1,200 @@
...
+                self._connection.client.terminate_session,
+                config=self._retry_config,
+                logger=_logger,
+                **request,
+            )
+        except Exception as e:
+            _logger.exception("Failed to terminate session.")
+            raise OperationalError(*e.args) from e
+
+    def __poll(self, query_id: str) -> Union[AthenaQueryExecution, AthenaCalculationExecution]:
```
Avinash-1394, line 131: Would this actually return `AthenaQueryExecution`?

pyathena/model.py
```diff
@@ -241,6 +241,158 @@ def result_reuse_minutes(self) -> Optional[int]:
...
+        return self._progress
+
+
+class AthenaCalculationExecution(AthenaCalculationExecutionStatus):
+    def __init__(self, response: Dict[str, Any]) -> None:
+        super(AthenaCalculationExecution, self).__init__(response)
+
+        self._calculation_id: Optional[str] = response.get("CalculationExecutionId", None)
+        if not self._calculation_id:
+            raise DataError("KeyError `CalculationExecutionId`")
```
Avinash-1394, line 300: suggestion (non-blocking): Might be just a preference but if would be good to put this piece of logic into a function and call it like self.get_calculation_id(response.get("CalculationExecutionId")) and raise the DataError in that function

pyathena/model.py
```diff
@@ -241,6 +241,158 @@ def result_reuse_minutes(self) -> Optional[int]:
...
+    @property
+    def progress(self) -> Optional[str]:
+        return self._progress
+
+
+class AthenaCalculationExecution(AthenaCalculationExecutionStatus):
+    def __init__(self, response: Dict[str, Any]) -> None:
+        super(AthenaCalculationExecution, self).__init__(response)
+
+        self._calculation_id: Optional[str] = response.get("CalculationExecutionId", None)
```
Avinash-1394, line 298: nitpick: You can simply leave this as `response.get("CalculationExecutionId")` since `.get` by default returns None if not found

pyathena/model.py
```diff
@@ -241,6 +241,158 @@ def result_reuse_minutes(self) -> Optional[int]:
         return self._result_reuse_minutes
 
 
+class AthenaCalculationExecutionStatus:
+    STATE_CREATING: str = "CREATING"
+    STATE_CREATED: str = "CREATED"
+    STATE_QUEUED: str = "QUEUED"
+    STATE_RUNNING: str = "RUNNING"
+    STATE_CANCELING: str = "CANCELING"
+    STATE_CANCELLED: str = "CANCELLED"
```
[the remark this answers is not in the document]
laughingman7743: Fixed: https://github.com/laughingman7743/PyAthena/pull/497/commits/a65dd478f8c754ce3b08ac2e6d8b404f96caa042

pyathena/model.py
laughingman7743: Fixed: https://github.com/laughingman7743/PyAthena/pull/497/commits/8a6202bd3a4a7f6c4fc1b4ba0ccee22eab756525

pyathena/calc_cursor.py
laughingman7743: Add spark package & Rename to SparkCursor: https://github.com/laughingman7743/PyAthena/pull/497/commits/f286cfe56f2c6c49f081fa2c63c1dcfaa7bacde9

pyathena/model.py
laughingman7743: I prefer to parse the value in the constructor of the current implementation, as it is harder to notice where the value is going wrong if there is logic in the get method.

pyathena/calc_cursor.py
```diff
@@ -0,0 +1,200 @@
...
+    def query_id(self, val: str) -> None:
+        self._calculation_id = val
+
+    @staticmethod
+    def get_default_engine_configuration() -> Dict[str, Any]:
+        return {
+            "CoordinatorDpuSize": 1,
+            "MaxConcurrentDpus": 2,
+            "DefaultExecutorDpuSize": 1,
+        }
```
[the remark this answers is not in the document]
laughingman7743: https://github.com/laughingman7743/PyAthena/pull/497/files#diff-cf1f78827b750a610d7b69b0f37937a7c2551c84d04b6b193472209e7094a5aeR31-R35
This is a default value setting, which you can set in the constructor argument according to your use case.

```
import textwrap
from pyathena import connect

conn = connect(work_group="spark-primary", cursor_class=SparkCursor)
with conn.cursor(
    engine_configuration={
        "CoordinatorDpuSize": 1,
        "MaxConcurrentDpus": 20,
        "DefaultExecutorDpuSize": 1,
        "AdditionalConfigs": {"string": "string"},
        "SparkProperties": {"string": "string"},
    }
) as cursor:
    cursor.execute(
        textwrap.dedent(
            """
            spark.sql("create database if not exists spark_demo_database")
            """
        )
    )
```

pyathena/calc_cursor.py
```diff
@@ -0,0 +1,200 @@
...
+        self._engine_configuration = (
+            engine_configuration
+            if engine_configuration
+            else self.get_default_engine_configuration()
+        )
+        self._notebook_version = notebook_version
+        self._session_description = description
+        self._session_idle_timeout_minutes = session_idle_timeout_minutes
+
+        self._session_id = session_id if session_id else self._start_session()
```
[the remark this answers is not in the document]
laughingman7743: If session_id is specified, I will check if the session is IDLE or not, and if IDLE, use the session, and if TERMINATED, DEGRADED, or FAILED, make an error.

pyathena/calc_cursor.py
laughingman7743: SparkCursor only returns AthenaCalculationExecution, but the parent class BaseCursor may return AthenaQueryExecution, so I specify two classes in Union.

pyathena/model.py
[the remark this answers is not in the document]
laughingman7743: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/athena/client/get_query_execution.html
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/athena/client/get_calculation_execution.html
👀 The query_execution is `CANCELLED` and the calculation_execution is `CANCELED`.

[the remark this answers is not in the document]
laughingman7743: Fixed: https://github.com/laughingman7743/PyAthena/pull/497/commits/0e451c54c928f71b2e8e611c6fc293416f7849ab

pyathena/calc_cursor.py
laughingman7743: Add spark package & Rename to SparkCursor: https://github.com/laughingman7743/PyAthena/pull/497/commits/f3d49beab4eecf90bda5017e1342e576ce44b6c4

pyathena/model.py
laughingman7743: Fixed: https://github.com/laughingman7743/PyAthena/pull/497/commits/6529f5dfb6dfee601321c7fa5c6b28d264ca3d14

pyathena/calc_cursor.py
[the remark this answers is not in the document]
laughingman7743: https://github.com/laughingman7743/PyAthena/pull/497/commits/6423399e5e722848ef208a35c6d428cb150c284d

laughingman7743:
TODO:
- [ ] Implement AsyncSparkCursor
- [ ] Unit tests
- [ ] Docs

pyathena/model.py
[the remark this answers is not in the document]
Avinash-1394: I really hope that wasn't intentional from AWS 🤣

PR #499, opened 2024-01-09
support json

clonylu:
This to support JSON in pyathena

Example use case 

```python
# table
class TestTable(Base):
    __tablename__ = "my_test_table"
    # this table has a column that is of Iceberg Array format
	my_test_column: Mapped[Union[list[str], None]] = mapped_column(ARRAY(String))


# query
def query_table(cursor):
	q = select(
            func.cast(TestTable.my_test_column, JSON).label(
                "my_test_column"
            )
        )
	x = q.compile(
            session.bind,
            session.bind.dialect,
            compile_kwargs={"render_postcompile": True},
        )
	df = cursor.execute(str(x), parameters=x.params, na_values=[]).as_pandas()
	return df
```


laughingman7743:
If you have a proposal for implementation, register the issue first and discuss it.

PR #513, opened 2024-02-11
Migrate package manager from pdm to hatch

laughingman7743:


laughingman7743:
```
Syncing environment plugin requirements
Environment `test.py3.8` is incompatible: cannot locate Python: 3.8
```
🤔

PR #536, opened 2024-04-23
Add support for UPDATE and MERGE operation in formatter

Oskar-Bonde:
Currently the formatter doesn't escape single quotes when the operation is MERGE or UPDATE. With this change MERGE and UPDATE update operations will be formatted with the _escape_presto instead of _escape_hive


laughingman7743 (approved):
Thanks! LGTM

Issue #528, opened 2024-03-18
[Warning] SADeprecationWarning: The dbapi() classmethod on dialect classes has been renamed to import_dbapi().


moshe-shelly:
I'm using the library via SQLDatabase (from langchain_community.utilities)

Getting the following warning:

SADeprecationWarning: The dbapi() classmethod on dialect classes has been renamed to import_dbapi().  Implement an import_dbapi() classmethod directly on class <class 'pyathena.sqlalchemy.rest.AthenaRestDialect'> to remove this warning; the old .dbapi() classmethod may be maintained for backwards compatibility.

I think this here is where the change should be implemented:
https://github.com/laughingman7743/PyAthena/blob/cb901047861cda020d481430d08699300e61831d/pyathena/sqlalchemy/base.py#L1016

Thanks!

laughingman7743:
https://github.com/laughingman7743/PyAthena/issues/496

Issue #496, opened 2024-01-02
SQLAlchemy dialect uses deprecated dbapi() method


0Itsuki0:
Getting the following warning while using create_engine(). 
It appears that this library is using a deprecated method from SQLAlchemy:

`SADeprecationWarning: The dbapi() classmethod on dialect classes has been renamed to import_dbapi().  Implement an import_dbapi() classmethod directly on class <class 'pyathena.sqlalchemy.rest.AthenaRestDialect'> to remove this warning; the old .dbapi() classmethod may be maintained for backwards compatibility.`

laughingman7743:
Please review the following issues and PRs.
- https://github.com/laughingman7743/PyAthena/issues/447
- https://github.com/laughingman7743/PyAthena/issues/456
- https://github.com/laughingman7743/PyAthena/pull/457
Required for compatibility with SQLAlchemy 1.4 series.

Issue #456, opened 2023-06-29
AttributeError: type object 'AthenaRestDialect' has no attribute 'dbapi'


dominik-reiner:
Hi,
this relates to https://github.com/laughingman7743/PyAthena/issues/447#issuecomment-1612552725.

I am running SQLAlchemy 1.4.48 and PyAthena 3.0.3.

When trying to create an engine as shown below I get an error:
AttributeError: type object 'AthenaRestDialect' has no attribute 'dbapi'

```
from sqlalchemy.engine import create_engine

conn_str = "awsathena+rest://{aws_access_key_id}:{aws_secret_access_key}@athena.{region_name}.amazonaws.com:443/"\
            "{schema_name}?s3_staging_dir={s3_staging_dir}"

engine = create_engine(conn_str.format(
  aws_access_key_id="XXX",
  aws_secret_access_key="XXX",
  region_name="eu-central-1",
  schema_name="default",
  s3_staging_dir="s3://XXX")
)
```

Here is the traceback:

---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[10], line 6
      1 from sqlalchemy.engine import create_engine
      3 conn_str = "awsathena+rest://{aws_access_key_id}:{aws_secret_access_key}@athena.{region_name}.amazonaws.com:443/"\
      4             "{schema_name}?s3_staging_dir={s3_staging_dir}"
----> 6 engine = create_engine(conn_str.format(
      7     aws_access_key_id="XXX",
      8     aws_secret_access_key="XXX",
      9     region_name="eu-central-1",
     10     schema_name="default",
     11     s3_staging_dir="s3://XXX"))

File :2, in create_engine(url, **kwargs)

File /opt/homebrew/Caskroom/miniforge/base/envs/xxx/lib/python3.9/site-packages/sqlalchemy/util/deprecations.py:375, in deprecated_params..decorate..warned(fn, *args, **kwargs)
    368     if m in kwargs:
    369         _warn_with_version(
    370             messages[m],
    371             versions[m],
    372             version_warnings[m],
    373             stacklevel=3,
    374         )
--> 375 return fn(*args, **kwargs)

File /opt/homebrew/Caskroom/miniforge/base/envs/xxx/lib/python3.9/site-packages/sqlalchemy/engine/create.py:541, in create_engine(url, **kwargs)
    539 if dbapi is None:
    540     dbapi_args = {}
[...]

Issue #447, opened 2023-05-30
PyAthena 3.0.3 + AthenaPandasDialect/AthenaRestDialect  has no attribute dbapi


mdeshmu:
With SQLAlchemy 1.4.x, getting below errors when using Athena Rest or Pandas dialect with PyAthena 3.0.3
`type object 'AthenaPandasDialect' has no attribute 'dbapi'`

Below is the stack trace.

2023-05-29 09:48:52,990:ERROR:flask_appbuilder.api: type object 'AthenaPandasDialect' has no attribute 'dbapi'
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/flask_appbuilder/api/__init__.py", line 110, in wraps
    return f(self, *args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/flask_appbuilder/api/__init__.py", line 182, in wraps
    return f(self, *args, **kwargs)
  File "/app/superset/views/base_api.py", line 122, in wraps
    raise ex
  File "/app/superset/views/base_api.py", line 113, in wraps
    duration, response = time_function(f, self, *args, **kwargs)
  File "/app/superset/utils/core.py", line 1586, in time_function
    response = func(*args, **kwargs)
  File "/app/superset/utils/log.py", line 266, in wrapper
    value = f(*args, **kwargs)
  File "/app/superset/databases/api.py", line 551, in schemas
    schemas = database.get_all_schema_names(
  File "/app/superset/utils/cache.py", line 131, in wrapped_f
    return f(*args, **kwargs)
  File "/app/superset/models/core.py", line 720, in get_all_schema_names
    raise self.db_engine_spec.get_dbapi_mapped_exception(ex) from ex
  File "/app/superset/models/core.py", line 717, in get_all_schema_names
    with self.get_inspector_with_context(ssh_tunnel=ssh_tunnel) as inspector:
  File "/usr/local/lib/python3.8/contextlib.py", line 113, in __enter__
    return next(self.gen)
  File "/app/superset/models/core.py", line 690, in get_inspector_with_context
    with self.get_sqla_engine_with_context(
  File "/usr/local/lib/python3.8/contextlib.py", line 113, in __enter__
    return next(self.gen)
  File "/app/superset/models/core.py", line 409, in get_sqla_engine_with_context
    yield self._get_sqla_engine(
  File "/app/superset/models/core.py", line 471, in _get_sqla_engine
[...]

PR #538, opened 2024-04-29
Updated dbapi

taradepan:
image
I fixed this warning.

### DIFF
--- a/pyathena/sqlalchemy/rest.py
+++ b/pyathena/sqlalchemy/rest.py
@@ -1,7 +1,12 @@
 # -*- coding: utf-8 -*-
 from pyathena.sqlalchemy.base import AthenaDialect
 
-
 class AthenaRestDialect(AthenaDialect):
     driver = "rest"
     supports_statement_cache = True
+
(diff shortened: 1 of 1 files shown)


laughingman7743:
The import_dbapi method is already defined in the parent class. The warnings are displayed by defining the following method to maintain compatibility with SQLAlchemy 1.4.
https://github.com/laughingman7743/PyAthena/blob/master/pyathena/sqlalchemy/base.py#L1011-L1013

If this method is deleted, the warning message will no longer be displayed, but you will not be able to use this library with SQLAlchemy 1.4.
https://github.com/laughingman7743/PyAthena/blob/master/pyathena/sqlalchemy/base.py#L1015-L1017

This issue has been addressed many times in the past. You may want to check past issues.
https://github.com/laughingman7743/PyAthena/issues/528
https://github.com/laughingman7743/PyAthena/issues/496
https://github.com/laughingman7743/PyAthena/issues/456
https://github.com/laughingman7743/PyAthena/issues/447

PR #557, opened 2024-08-17
Support Athena parameterized queries when paramstyle is qmark (fix #545)

### LINKED ISSUE #545: Feature request: support for positional parameter substitution
Thank you for a wonderful library and for your work maintaining it!

One feature I hope can be considered for future development is implementing `?` style param support. Athena supports [parameterized queries](https://docs.aws.amazon.com/athena/latest/ug/querying-with-prepared-statements.html) using the `?` character for positional substitution.

For use cases where queries are run both using `pyathena` and by analysts directly, having `PyFormat`-style params is not ideal as it requires analysts to modify queries to execute them directly in their SQL clients or in the Athena console. Supporting `?` for query parameterization would decouple the query from any specific execution context, making it more portable.

Thanks for your time and consideration.

laughingman7743:


laughingman7743:
TODO:
- [ ] Unit tests

PR #577, opened 2025-04-06
Allow the use of use_fips_endpoint arg for client connections

avivg37:
`boto3.client(..., use_fips_endpoint=True)` tells the Boto3 client to prioritize using FIPS 140-2 compliant endpoints.

The `use_fips_endpoint` parameter in the Python AWS client (Boto3) setup controls whether the client will communicate with Federal Information Processing Standards (FIPS) 140-2 compliant endpoints.
FIPS endpoints are AWS service endpoints that utilize FIPS 140-2 validated cryptographic modules for encrypting data in transit. This ensures a higher level of security and compliance for sensitive workloads.
If your application needs to comply with FIPS 140-2 standards (often required for U.S. federal government agencies and organizations working with them), you must use FIPS endpoints.

This PR allows this parameter to be set when needed.

See: `Amazon Athena` in https://aws.amazon.com/compliance/fips/

### DIFF
--- a/pyathena/connection.py
+++ b/pyathena/connection.py
@@ -60,6 +60,7 @@ class Connection(Generic[ConnectionCursor]):
         "endpoint_url",
         "region_name",
         "config",
+        "use_fips_endpoint",
     ]
 
     @overload


laughingman7743:
Thank you for your contribution.
I checked the botocore documentation and it seems that `use_fips_endpoint` is set in botocore.config.
https://botocore.amazonaws.com/v1/documentation/api/latest/reference/config.html
It is not possible to set it as an argument for the boto3.session object.
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/session.html
The botocore.config object can be specified as follows and should be passed to the boto3.session object.
```
from botocore.config import Config
from pyathena import connect

conn = connect(s3_staging_dir="s3://YOUR_S3_BUCKET/path/to/",
    region_name="us-west-2",
    ....
    config=Config(
        use_fips_endpoint=True,
        ...
    )
)
```
Please check it.

Issue #602, opened 2025-09-03
Reserved words are not quoted correctly


izzudinhafiz:
When using PyAthena with Sqlalchemy, reserved keywords aren't being quoted correctly. The test case is also failing to catch this as it's compiling with the default dialect instead of PyAthena's Dialect

I have a minimum reproducible snippet here

```python
from pyathena.sqlalchemy.base import AthenaDialect
from sqlalchemy import Column, MetaData, String, Table

dialect = AthenaDialect()

# this is from the test code
test_case = Table("select", MetaData(), Column("current_timestamp", String()))
should_quote = Table("select", MetaData(), Column("current_timestamp", String()))

test_case_query = str(test_case.select().where(test_case.c.current_timestamp == "a"))
should_quote_query = (
    should_quote.select()
    .where(should_quote.c.current_timestamp == "a")
    .compile(dialect=dialect)
    .string
)

assert '"select"' in test_case_query, test_case_query
assert '"select"' in should_quote_query, should_quote_query
```

This will fail the assertion on should_quote_query

The query from the test case outputs to 
```sql
SELECT "select"."current_timestamp" 
FROM "select" 
WHERE "select"."current_timestamp" = :current_timestamp_1
```

Where as when using the dialect
```sql
SELECT select.current_timestamp 
FROM select 
WHERE select.current_timestamp = %(current_timestamp_1)s
```

The impacted test_case is here 

```python
class TestSQLAlchemyAthena:
    def test_reserved_words(self):
        """Presto uses double quotes, not backticks"""
        fake_table = Table("select", MetaData(), Column("current_timestamp", types.String()))
        query = str(fake_table.select().where(fake_table.c.current_timestamp == "a"))
        assert '"select"' in query
        assert '"current_timestamp"' in query
        assert "`select`" not in query
        assert "`current_timestamp`" not in query
```

I'll be opening a PR to fix this issue

izzudinhafiz:
Issue coming from `sqlalchemy.sql.compiler.IdentifierPreparer._requires_quote` function which downcases the value before checking against the reserved_words

```python
class IdentifierPreparer:
    def _requires_quotes(self, value: str) -> bool:
        """Return True if the given identifier requires quoting."""
        lc_value = value.lower()
        return (
            lc_value in self.reserved_words
            or value[0] in self.illegal_initial_characters
            or not self.legal_characters.match(str(value))
            or (lc_value != value)
        )
```

The reserved words in the default dialect are in lower case

```python
[...]

PR #603, opened 2025-09-03
Fix: Reserved words are not quoted correctly

izzudinhafiz:
Re: #602 

I’m not able to run the test locally. Is there a way to run it in CI?


laughingman7743 (approved):
## ✅ Code Review: LGTM (Looks Good To Me)

### Summary
This PR correctly fixes the issue where reserved words were not being quoted properly in SQLAlchemy queries. The root cause was identified accurately - SQLAlchemy's  performs case-sensitive matching against the  set, so uppercase reserved words wouldn't match lowercase identifiers used in actual SQL.

### Key Improvements

1. **Correct Root Cause Analysis**: Converting all reserved words from uppercase to lowercase ensures proper matching with actual SQL identifiers
2. **Comprehensive Fix**: Applied to both  and 
3. **Enhanced Test Coverage**: 
   - Uses Athena-specific reserved words (, ) that are not reserved in the default dialect
   - Explicitly compiles with Athena dialect instead of relying on string conversion
   - Much more reliable test that would catch regressions

### Test Results ✅
- **Reserved words test**: 
- **Lint check (ruff)**: 
- **Format check**:   
- **Type check (mypy)**: 
- **Other SQLAlchemy tests**: Running successfully without regressions

### Technical Validation
The changes are mechanically safe (simple case conversion) and align with real-world SQL usage patterns where identifiers are typically lowercase.

**Recommendation**: Ready to merge! This fix will make SQLAlchemy integration more robust by ensuring Athena reserved words are properly quoted.

laughingman7743 (approved):
LGTM

PR #626, opened 2025-11-10
trailing slash check for s3_staging_dir

CAinDE:
If the trailing slash isn't there, Athena throws a generic error:

[ErrorCode: INTERNAL_ERROR_QUERY_ENGINE] Amazon Athena experienced an internal error while executing this query. Please contact AWS support for further assistance. You will not be charged for this query. We apologize for the inconvenience.

### DIFF
--- a/pyathena/connection.py
+++ b/pyathena/connection.py
@@ -261,7 +261,10 @@ class Connection(Generic[ConnectionCursor]):
         assert self.s3_staging_dir or self.work_group, (
             "Required argument `s3_staging_dir` or `work_group` not found."
         )
-
+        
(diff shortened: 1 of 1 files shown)


laughingman7743:
Thank you for your contribution.
To ensure this change works effectively not only for SQLAlchemy, I think it would be good to add it to the following locations.
https://github.com/laughingman7743/PyAthena/blob/master/pyathena/connection.py#L264

CAinDE:
That totally makes sense :) I updated it.

laughingman7743:
Thanks!

PR #740, opened 2026-07-10
Fix/739 cache schema catalog

### LINKED ISSUE #739: PyAthena caching does not check schema or catalog names
It appears that when checking for a query for an earlier cached value, PyAthena checks the string query for equality, but it doesn't appear to check query execution context (the catalog or the database) is the same as that requested. Ideally, it would only reuse the cache if the query and its context matched.

I hit this because I have a dev-and-prod setup within Athena where the same tables exist in both dev and prod environments. If I attempt to use the dev database for the same query as an earlier prod query, I get the prod results, not my dev results, which is quite confusing.

Workaround of course is to disable caching for any such queries, which is fine, but obviously not ideal.

Liam3851:
## WHAT
Check the database name and the catalog are the same for a request before allowing a query cache hit. See #739. 

## WHY
When the pyathena cache is checked for hits, queries from the given workgroup are checked for having the same query string, but NOT whether they were queried against the same Catalog or DatabaseName. Thus queries among multiple environments with the same table names and schemas (e.g., dev vs. prod catalogs or databases) will cause spurious cache hits from the other environment, possibly with erroneous data.

### DIFF
--- a/pyathena/aio/common.py
+++ b/pyathena/aio/common.py
@@ -220,8 +220,8 @@ class AioBaseCursor(BaseCursor):
                         break
                     if (
                         execution.query == query
-                        and (self._schema_name is None or execution.database == self._schema_name)
-                        and (self._catalog_name is None or execution.catalog == self._catalog_name)
+                        and execution.database == self._schema_name
+                        and execution.catalog == self._catalog_name
                     ):
                         query_id = execution.query_id
                         break
--- a/pyathena/aio/common.py
+++ b/pyathena/aio/common.py
@@ -221,7 +221,7 @@ class AioBaseCursor(BaseCursor):
                     if (
                         execution.query == query
                         and execution.database == self._schema_name
-                        and execution.catalog == self._catalog_name
+                        and (execution.catalog or "").lower() == (self._catalog_name or "").lower()
                     ):
                         query_id = execution.query_id
                         break
--- a/pyathena/aio/common.py
+++ b/pyathena/aio/common.py
@@ -218,7 +218,11 @@ class AioBaseCursor(BaseCursor):
                     ):
                         next_token = None
                         break
-                    if execution.query == query:
+                    if (
+                        execution.query == query
+                        and (self._schema_name is None or execution.database == self._schema_name)
+                        and (self._catalog_name is None or execution.catalog == self._catalog_name)
+                    ):
                         query_id = execution.query_id
                         break
                 if query_id or next_token is None:
--- a/pyathena/common.py
+++ b/pyathena/common.py
@@ -638,7 +638,11 @@ class BaseCursor(metaclass=ABCMeta):
                     ):
                         next_token = None
                         break
-                    if execution.query == query:
+                    if (
+                        execution.query == query
+                        and (self._schema_name is None or execution.database == self._schema_name)
+                        and (self._catalog_name is None or execution.catalog == self._catalog_name)
+                    ):
                         query_id = execution.query_id
                         break
                 if query_id or next_token is None:
--- a/pyathena/common.py
+++ b/pyathena/common.py
@@ -641,7 +641,7 @@ class BaseCursor(metaclass=ABCMeta):
                     if (
                         execution.query == query
                         and execution.database == self._schema_name
-                        and execution.catalog == self._catalog_name
+                        and (execution.catalog or "").lower() == (self._catalog_name or "").lower()
                     ):
                         query_id = execution.query_id
                         break
--- a/pyathena/common.py
+++ b/pyathena/common.py
@@ -640,8 +640,8 @@ class BaseCursor(metaclass=ABCMeta):
                         break
                     if (
                         execution.query == query
-                        and (self._schema_name is None or execution.database == self._schema_name)
-                        and (self._catalog_name is None or execution.catalog == self._catalog_name)
+                        and execution.database == self._schema_name
+                        and execution.catalog == self._catalog_name
                     ):
                         query_id = execution.query_id
                         break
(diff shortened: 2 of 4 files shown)


pyathena/common.py
```diff
@@ -638,7 +638,11 @@ def _find_previous_query_id(
                     ):
                         next_token = None
                         break
-                    if execution.query == query:
+                    if (
+                        execution.query == query
+                        and execution.database == self._schema_name
+                        and execution.catalog == self._catalog_name
```
laughingman7743: Thanks for the fix — the direction is exactly right. One issue with the catalog comparison, though: Athena normalizes catalog names to lowercase in query execution records. I verified this against live Athena: submitting `Catalog: "AwsDataCatalog"` to `StartQueryExecution` comes back as `Catalog: "awsdatacatalog"` from `GetQueryExecution`.

Since this line compares the recorded (lowercased) value against the user-supplied connection value, anyone connecting with `catalog_name="AwsDataCatalog"` (the casing used throughout the AWS docs) would never get a cache hit again — even for queries issued by the same connection. That's a silent caching regression.

A case-insensitive comparison is safe here: catalog names are case-insensitive in Athena. I also verified that creating a data catalog whose name differs only in case from an existing one fails with `InvalidRequestException: DataCatalog ... has already been created`, so two distinct catalogs can never differ only by case. Something like:

```python
and (execution.catalog or "").lower() == (self._catalog_name or "").lower()
```

This keeps the strict `None`-vs-value semantics (a recorded execution without a catalog still won't match a connection with one, which is the safe default) while fixing the case mismatch.

The database comparison is fine as is — Athena echoes the submitted `Database` value verbatim (no normalization), so queries issued through the same connection always match.

pyathena/aio/common.py
```diff
@@ -218,7 +218,11 @@ async def _find_previous_query_id(  # type: ignore[override]
                     ):
                         next_token = None
                         break
-                    if execution.query == query:
+                    if (
+                        execution.query == query
+                        and execution.database == self._schema_name
+                        and execution.catalog == self._catalog_name
```
laughingman7743: Same catalog case-sensitivity issue as in `pyathena/common.py` — see the comment there. Athena records catalog names lowercased, so this comparison should be case-insensitive as well.

tests/pyathena/aio/test_cursor.py
```diff
@@ -130,6 +130,48 @@ async def test_execute_internal_legacy_kwargs_passthrough(self):
...
+            return AthenaQueryExecution(
+                {
+                    "QueryExecution": {
+                        "QueryExecutionId": f"query_id_{schema}",
+                        "Query": query,
+                        "StatementType": AthenaQueryExecution.STATEMENT_TYPE_DML,
+                        "QueryExecutionContext": {"Database": schema},
+                        "Status": {
+                            "State": AthenaQueryExecution.STATE_SUCCEEDED,
+                            "CompletionDateTime": datetime.now(),
```
laughingman7743: Nit: the synchronous counterpart in `tests/pyathena/test_cursor.py` uses `datetime.now(timezone.utc)` — please use the timezone-aware form here too for consistency. `_find_previous_query_id` sorts executions by `completion_date_time`, so mixing naive and aware datetimes in a future test extension would raise a `TypeError`.

tests/pyathena/test_cursor.py
```diff
@@ -185,6 +185,44 @@ def test_cache_expiration_time_with_cache_size(self, cursor):
...
+                        "Status": {
+                            "State": AthenaQueryExecution.STATE_SUCCEEDED,
+                            "CompletionDateTime": datetime.now(timezone.utc),
+                        },
+                    }
+                }
+            )
+
+        cursor = Cursor.__new__(Cursor)  # bypass __init__ to avoid AWS calls
+        cursor._catalog_name = None
```
laughingman7743, line 213: The test approach itself looks good — it follows the same `__new__` + `patch.object` pattern as the existing no-AWS unit tests in this file, and the negative + positive assertion pair guards against `_find_previous_query_id`'s broad `except Exception` masking a setup mistake. I confirmed both new tests pass locally on this branch.

One coverage gap: with `_catalog_name = None` and no `Catalog` in the mocked execution context, the new catalog condition is only ever exercised as `None == None`, so the catalog half of the fix is untested. Could you extend this test (and the aio counterpart) with a catalog-mismatch case — e.g. an execution recorded with `"Catalog": "some_catalog"` while the cursor has a different `_catalog_name`? If the case-insensitive comparison suggested in `pyathena/common.py` is adopted, a case-difference case (`"awsdatacatalog"` recorded vs. `"AwsDataCatalog"` on the cursor, expecting a hit) would pin that behavior down too.

laughingman7743:
A note on the failed CI run: the failure is environmental, not caused by this PR. The test workflow authenticates to AWS via OIDC, and GitHub does not issue OIDC tokens to `pull_request` runs triggered from forks, so `configure-aws-credentials` fails before any tests run. This is expected for fork PRs against this repo.

I've verified this branch locally instead: both new unit tests pass, and `just lint` (ruff + format check + mypy) is clean. Once the review comments are addressed, I'll run the full integration test suite against AWS on my side before merging.

Liam3851:
Thanks for the review @laughingman7743 (and catching the case sensitivity issue); updated the PR.

laughingman7743 (approved):
LGTM — all review comments addressed: catalog comparison is now case-insensitive with strict None semantics, and the new catalog mismatch / case-difference tests cover both directions.

Verified locally on the final branch: all 4 new unit tests pass, `just lint` (ruff + format + mypy) is clean, and the existing cache integration tests (`-k cache`) pass against live Athena — confirming real-world cache hits still work with the new comparison. Thanks for the fix!

PR #743, opened 2026-07-25
fix(sqlalchemy): return the formatted DATE literal from AthenaDate.process

### LINKED ISSUE #742: AthenaDate.process drops the formatted DATE literal (missing return)
## Description

`AthenaDate.process` in `pyathena/sqlalchemy/types.py` builds a `DATE 'YYYY-MM-DD'` literal for `date`/`datetime` values, but the formatted string in the `if` branch is never returned:

```python
@staticmethod
def process(value: date | datetime | Any) -> str:
    if isinstance(value, (date, datetime)):
        f"DATE '{value:%Y-%m-%d}'"   # computed and discarded — no return
    return f"DATE '{value!s}'"
```

The `if` branch evaluates the f-string and throws it away, so every value falls through to `return f"DATE '{value!s}'"`.

## Impact

This is the `literal_processor` used when an `AthenaDate` value is rendered inline (e.g. with `literal_binds`).

- For a `date` value, `str(value)` is `YYYY-MM-DD`, so the output is correct by accident.
- For a `datetime` value, `str(value)` is `YYYY-MM-DD HH:MM:SS`, so it renders `DATE '2017-01-01 12:00:00'`, which is not a valid Athena `DATE` literal.

For comparison, `pyathena/formatter.py` renders it correctly:

```python
def _format_date(formatter, escaper, val):
    return f"DATE '{val:%Y-%m-%d}'"
```

## Fix

Return the formatted literal from the `if` branch:

```python
if isinstance(value, (date, datetime)):
    return f"DATE '{value:%Y-%m-%d}'"
return f"DATE '{value!s}'"
```

Sanjays2402:
Closes #742

## WHAT
`AthenaDate.process` now returns the `DATE 'YYYY-MM-DD'` literal it builds in the `date`/`datetime` branch, instead of discarding it and falling through to `DATE '{value!s}'`. Adds a regression test covering both a `date` and a `datetime` input.

## WHY
The `if` branch evaluated the f-string without returning it, so a `datetime` rendered as `DATE '2017-01-01 12:00:00'` — not a valid Athena `DATE` literal. `pyathena/formatter.py` already formats it correctly.

### DIFF
--- a/pyathena/sqlalchemy/types.py
+++ b/pyathena/sqlalchemy/types.py
@@ -81,7 +81,7 @@ class AthenaDate(TypeEngine[date]):
     @staticmethod
     def process(value: date | datetime | Any) -> str:
         if isinstance(value, (date, datetime)):
-            f"DATE '{value:%Y-%m-%d}'"
+            return f"DATE '{value:%Y-%m-%d}'"
         return f"DATE '{value!s}'"
 
     def literal_processor(self, dialect: Dialect) -> _LiteralProcessorType[date] | None:
--- a/tests/pyathena/sqlalchemy/test_types.py
+++ b/tests/pyathena/sqlalchemy/test_types.py
@@ -1,3 +1,5 @@
+from datetime import date, datetime
+
 import pytest
 from sqlalchemy import Integer, String, types
 from sqlalchemy.sql import sqltypes
@@ -7,6 +9,7 @@ from pyathena.sqlalchemy.types import (
     MAP,
     STRUCT,
     AthenaArray,
+    AthenaDate,
     AthenaMap,
     AthenaStruct,
     get_double_type,
@@ -163,3 +166,15 @@ def test_get_double_type():
     else:
         assert result is types.FLOAT
     assert ischema_names["double"] is result
+
+
+class TestAthenaDate:
+    @pytest.mark.parametrize(
+        ("value", "expected"),
+        [
+            (date(2017, 1, 1), "DATE '2017-01-01'"),
+            (datetime(2017, 1, 1, 12, 34, 56), "DATE '2017-01-01'"),
+        ],
+    )
+    def test_process_renders_date_only_literal(self, value, expected):
+        assert AthenaDate.process(value) == expected


pyathena/sqlalchemy/types.py
```diff
@@ -81,7 +81,7 @@ class AthenaDate(TypeEngine[date]):
     @staticmethod
     def process(value: date | datetime | Any) -> str:
         if isinstance(value, (date, datetime)):
```
laughingman7743, line 83: Nit, optional and out of scope for this PR: `datetime` is a subclass of `date`, so the tuple is redundant — `isinstance(value, date)` matches both. The `date | datetime | Any` annotation on the line above collapses to `Any` for the same reason. Both predate this change, so no need to touch them here; just noting it for a future cleanup.

tests/pyathena/sqlalchemy/test_types.py
```diff
@@ -163,3 +166,15 @@ def test_get_double_type():
...
+class TestAthenaDate:
+    @pytest.mark.parametrize(
+        ("value", "expected"),
+        [
+            (date(2017, 1, 1), "DATE '2017-01-01'"),
+            (datetime(2017, 1, 1, 12, 34, 56), "DATE '2017-01-01'"),
+        ],
+    )
+    def test_process_renders_date_only_literal(self, value, expected):
+        assert AthenaDate.process(value) == expected
```
laughingman7743, line 180: Nice — parametrizing over both `date` and `datetime` is exactly the right shape here, and grouping it under a `Test<ClassName>` class matches the convention in this repo. I verified locally that reverting the `return` makes the `datetime` case fail with `DATE '2017-01-01 12:34:56'`, so this is a real regression guard rather than a tautology.

Optional follow-up (happy to take it as a separate PR, or from anyone else): the sibling `AthenaTimestamp.process` has no unit test. It is currently correct — it does have its `return` — but nothing guards it. No need to expand this PR for it.

laughingman7743 (approved):

PR #752, opened 2026-08-11
Add a worktree environment helper

laughingman7743:
## WHAT

Add a worktree-env script that links the main checkout .env file into the current git worktree.
Expose the script through the just worktree-env recipe and document the workflow in AGENTS.md.

## WHY

Gitignored environment files are not copied when git creates a worktree.
Without the link, AWS integration tests fail before collection even though the main checkout already has the required configuration.

The helper derives the main checkout from the git common directory, refuses to replace a regular .env file, and makes no change when run from the main checkout.

## TESTING

- bash syntax validation
- just worktree-env, including an idempotent second run
- Verified that the generated link remains gitignored.
- Verified that execution from the main checkout makes no change.
- just lint
- just docs lint

### DIFF
--- a/justfile
+++ b/justfile
@@ -7,7 +7,7 @@ default:
 
 # Link the main checkout's gitignored .env into a worktree
 worktree-env:
-    bash scripts/worktree-env.sh
+    scripts/worktree-env.sh
 
 # Auto-fix formatting and imports
 format:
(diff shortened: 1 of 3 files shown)


justfile
```diff
@@ -5,6 +5,10 @@ TOX_VERSION := "4.34.1"
 default:
     @just --list
 
+# Link the main checkout's gitignored .env into a worktree
+worktree-env:
+    bash scripts/worktree-env.sh
```
laughingman7743:
```suggestion
    scripts/worktree-env.sh
```

PR #763, opened 2026-09-01
Fall back to the original string for complex MAP/STRUCT/ARRAY values instead of returning None

### LINKED ISSUE #764: Complex MAP/STRUCT/ARRAY values are silently converted to None (looks like data loss)
## Describe the bug

When a complex-type column contains values that are too complex for the native-format parsers, `_to_array` / `_to_map` / `_to_struct` return `None`. To the caller this is indistinguishable from a NULL cell, so the value appears **silently lost** even though Athena returned it.

## To reproduce

```python
from pyathena.converter import _to_map

_to_map('{push=Y}')
# {'push': 'Y'}  -> displayed fine

_to_map('{items=[{"product_id":285,"option_id":6049}], url=/webview/checkout, brand_id=75}')
# None  -> displayed as NULL, data appears lost
```

Any `MAP<VARCHAR, VARCHAR>` whose value contains `[](){}` (e.g. a serialized JSON array) hits this path in `_to_map`:

```python
if any(char in inner for char in "()[]"):
    # Contains complex structures (arrays, structs), skip parsing
    return None
```

## Expected behavior

The existing comment in `_to_map` already states the intent:

```python
# But for complex structures, return None to keep as string
```

…but returning `None` does not keep the string. When the parser gives up, the original varchar value should be returned so the data is preserved (as a string), matching the comment's intent.

## Real-world impact

In BI tools built on PyAthena (e.g. Apache Superset), simple maps like `{push=Y}` display fine while complex maps display as NULL. Users reasonably conclude the data pipeline dropped their data — we spent a full debugging session tracing Kafka → Flink → Iceberg storage before finding the converter. `CAST(col AS JSON)` works as a workaround, but the silent NULL is very misleading.

## Environment

- PyAthena: 3.35.4 (also reproduced on current master)
- Python: 3.14

seunggabi:
Closes #764

## Problem

When a complex-type column contains values that are too complex for the native-format parsers, `_to_array` / `_to_map` / `_to_struct` return `None`. To the caller this is indistinguishable from a NULL cell, so the data appears **silently lost**.

Real-world example: a `MAP<VARCHAR, VARCHAR>` column where one value holds a serialized JSON array:

```
{items=[{"product_id":285,"option_id":6049}], url=/webview/checkout, brand_id=75}
```

```python
_to_map('{push=Y}')                      # {'push': 'Y'}  -> shown
_to_map('{items=[{...}], brand_id=75}')  # None           -> shown as NULL
```

In BI tools built on PyAthena (e.g. Superset), simple maps display fine while complex ones display as NULL — even though Athena returned the value. Users reasonably conclude the data pipeline dropped their data. (We spent a full debugging session tracing Kafka → Flink → Iceberg before finding the converter.)

Notably, the existing comment in `_to_map` already states the intent:

```python
# But for complex structures, return None to keep as string
```

…but returning `None` does not keep the string.

## Change

On the give-up paths (nested-structure skip and parse exceptions), return the **original varchar value** instead of `None`:

- `_to_array`: nested-array skip / parse exception → original string
- `_to_map`: complex-structure skip / parse exception → original string
- `_to_struct`: parse exception → original string
- `parser.py`: widened `struct_parser` type annotation accordingly

Behavior is unchanged for:
- `None` input → `None`
- values that don't look like the type at all → `None`
- all simple/parsable cases → parsed `dict` / `list` as before

`CAST(col AS JSON)` remains the recommended way to get reliably structured values; this change only ensures the fallback is lossless.

## Tests

- Added tests for the raw-string fallback (map with nested array values incl. the real-world case above, native nested arrays)
- Added regression tests that simple values still parse (`{push=Y}`)
- Updated 2 existing expectations in `test_to_array_complex_nested_cases` from `None` to the original string
- `pytest tests/pyathena/test_parser.py tests/pyathena/test_converter.py`: 130 passed
- `ruff check` / `ruff format --check` / `mypy`: clean

### DIFF
--- a/pyathena/converter.py
+++ b/pyathena/converter.py
@@ -89,7 +89,7 @@ def _to_json(varchar_value: str | None) -> Any | None:
     return json.loads(varchar_value)
 
 
-def _to_array(varchar_value: str | None) -> list[Any] | None:
+def _to_array(varchar_value: str | None) -> list[Any] | str | None:
     """Convert array data to Python list.
 
     Supports two formats:
@@ -102,7 +102,9 @@ def _to_array(varchar_value: str | None) -> list[Any] | None:
         varchar_value: String representation of array data
 
     Returns:
-        List representation of array, or None if parsing fails
+        List representation of array, the original string if the value is
+        too complex to parse (so the data is preserved), or None if the
+        input is None or does not look like an array
     """
     if varchar_value is None:
         return None
@@ -127,15 +129,16 @@ def _to_array(varchar_value: str | None) -> list[Any] | None:
     try:
         # For nested arrays, too complex for basic parsing
         if "[" in inner:
-            # Contains nested arrays - too complex for basic parsing
-            return None
+            # Contains nested arrays - keep the original string
+            # instead of silently dropping the value
+            return varchar_value
         # Try native parsing (including struct arrays)
         return _parse_array_native(inner)
     except Exception:
-        return None
+        return varchar_value
 
 
-def _to_map(varchar_value: str | None) -> dict[str, Any] | None:
+def _to_map(varchar_value: str | None) -> dict[str, Any] | str | None:
     """Convert map data to Python dictionary.
 
     Supports two formats:
@@ -148,7 +151,9 @@ def _to_map(varchar_value: str | None) -> dict[str, Any] | None:
         varchar_value: String representation of map data
 
     Returns:
-        Dictionary representation of map, or None if parsing fails
+        Dictionary representation of map, the original string if the value
+        is too complex to parse (so the data is preserved), or None if the
+        input is None or does not look like a map
     """
     if varchar_value is None:
         return None
@@ -177,16 +182,17 @@ def _to_map(varchar_value: str | None) -> dict[str, Any] | None:
 
     try:
         # MAP format is always key=value pairs
-        # But for complex structures, return None to keep as string
+        # But for complex structures, keep the original string
         if any(char in inner for char in "()[]"):
             # Contains complex structures (arrays, structs), skip parsing
-            return None
+            # and keep the original string instead of silently dropping it
+            return varchar_value
         return _parse_map_native(inner)
     except Exception:
-        return None
+        return varchar_value
 
 
-def _to_struct(varchar_value: str | None) -> dict[str, Any] | None:
+def _to_struct(varchar_value: str | None) -> dict[str, Any] | str | None:
     """Convert struct data to Python dictionary.
 
     Supports two formats:
@@ -199,7 +205,9 @@ def _to_struct(varchar_value: str | None) -> dict[str, Any] | None:
         varchar_value: String representation of struct data
 
     Returns:
-        Dictionary representation of struct, or None if parsing fails
+        Dictionary representation of struct, the original string if the
+        value is too complex to parse (so the data is preserved), or None
+        if the input is None or does not look like a struct
     """
     if varchar_value is None:
         return None
@@ -233,7 +241,8 @@ def _to_struct(varchar_value: str | None) -> dict[str, Any] | None:
         # Unnamed struct: {Alice, 25}
         return _parse_unnamed_struct(inner)
     except Exception:
-        return None
+        # Keep the original string instead of silently dropping the value
+        return varchar_value
 
 
 def _parse_array_native(inner: str) -> list[Any] | None:
--- a/pyathena/converter.py
+++ b/pyathena/converter.py
@@ -132,8 +132,10 @@ def _to_array(varchar_value: str | None) -> list[Any] | str | None:
             # Contains nested arrays - keep the original string
             # instead of silently dropping the value
             return varchar_value
-        # Try native parsing (including struct arrays)
-        return _parse_array_native(inner)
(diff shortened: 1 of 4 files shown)


pyathena/converter.py
```diff
@@ -177,16 +182,17 @@ def _to_map(varchar_value: str | None) -> dict[str, Any] | None:
...
     try:
         # MAP format is always key=value pairs
-        # But for complex structures, return None to keep as string
+        # But for complex structures, keep the original string
         if any(char in inner for char in "()[]"):
             # Contains complex structures (arrays, structs), skip parsing
-            return None
+            # and keep the original string instead of silently dropping it
+            return varchar_value
         return _parse_map_native(inner)
```
laughingman7743: Thanks for the fix — the direction matches the comment's original intent. One class of silent `None` still slips through here, though: values that are nested structs/maps (braces only, no `[]()`) pass the `"()[]"` check above, and `_parse_map_native` then skips every pair whose value contains `{}="` and returns `None` when nothing is left:

```python
_to_map('{a={b=1}}')       # None — e.g. MAP<VARCHAR, ROW(...)> or MAP<VARCHAR, MAP(...)>
_to_map('{a={b=1, c=2}}')  # None
```

Since these are exactly the kind of complex values this PR is about, could you extend the fallback to cover the helper's give-up path as well? For example:

```python
result = _parse_map_native(inner)
return result if result is not None else varchar_value
```

The same pattern applies to `_to_array` and `_to_struct` — see the other comments.

pyathena/converter.py
```diff
@@ -127,15 +129,16 @@ def _to_array(varchar_value: str | None) -> list[Any] | None:
     try:
         # For nested arrays, too complex for basic parsing
         if "[" in inner:
-            # Contains nested arrays - too complex for basic parsing
-            return None
+            # Contains nested arrays - keep the original string
+            # instead of silently dropping the value
+            return varchar_value
         # Try native parsing (including struct arrays)
         return _parse_array_native(inner)
```
laughingman7743: Same remaining gap here: `_parse_array_native` returns `None` when every item is skipped by its safety check (items containing `[]="`), and that `None` still propagates to the caller:

```python
_to_array('[a=1, b=2]')  # None — items containing '=' are skipped
```

Suggest the same treatment as in `_to_map`:

```python
result = _parse_array_native(inner)
return result if result is not None else varchar_value
```

pyathena/converter.py
```diff
@@ -233,7 +241,8 @@ def _to_struct(varchar_value: str | None) -> dict[str, Any] | None:
         # Unnamed struct: {Alice, 25}
         return _parse_unnamed_struct(inner)
```
laughingman7743, line 249: `_parse_named_struct` (the branch just above) can also return `None` when all pairs are skipped, which still propagates as a silent `None`:

```python
_to_struct('{"a"=1}')  # None — the key containing a quote is skipped
```

Worth applying the same `None` → original-string fallback to its result as well (`_parse_unnamed_struct` always returns a dict, so only the named branch needs it).

pyathena/converter.py
laughingman7743, line 1: One more case in the same class of silent data loss (pre-existing, but I'd like it folded into this PR since it directly undermines the "never silently lose data" goal): `_parse_map_native` silently drops individual pairs whose key/value contains `{}="`, so a partially parseable map returns a **partial dict** — which is arguably worse than `None` was, because the loss is invisible:

```python
_to_map('{a="x", b=1}')  # {'b': '1'} — the a="x" pair is silently dropped
```

`_parse_array_native` has the same issue for skipped items (`_to_array('[a, b=1]')` → `['a']`).

A clean way to handle this together with the other comments: make the `_parse_*_native` helpers return `None` whenever they had to skip anything, and let the outer functions fall back to the original string. That way the result is always either fully parsed or the intact raw string, never a partial value.

pyathena/converter.py
seunggabi: Good catch — fixed in f04125b. `_to_map` now falls back to the original string when `_parse_map_native` returns `None`, and added tests for `{a={b=1}}` / `{a={b=1, c=2}}`.

pyathena/converter.py
seunggabi: Fixed in f04125b — same `None` → original-string fallback applied to `_parse_array_native`'s result, with tests for `[a=1, b=2]`.

pyathena/converter.py
seunggabi: Fixed in f04125b — `_parse_named_struct`'s result now gets the same fallback (left `_parse_unnamed_struct` as-is since it always returns a dict), with a test for `{"a"=1}`.

pyathena/converter.py
seunggabi: Agreed that a partial dict/list is worse than `None` was — implemented exactly as you suggested in f04125b: `_parse_map_native` / `_parse_array_native` / `_parse_named_struct` now return `None` whenever any pair/item has to be skipped, and the outer functions fall back to the original string. The result is always either fully parsed or the intact raw string, never partial. Added tests for `{a="x", b=1}`, `{a, b=1}`, and `[a, b=1]` (previously `{'b': '1'}` / `['a']`). All 137 unit tests pass, ruff/mypy clean.

Issue #755, opened 2026-08-11
SQLAlchemy compliance: enable native ARRAY type support


laughingman7743:
## Context

Parent issue: #736

Athena supports native `array<T>` values, but the SQLAlchemy compliance requirement `array_type` is closed.
The dialect must expose ARRAY columns and values through a real SQLAlchemy array type before the requirement can be opened.

## Scope

- Audit the current ARRAY type mapping, reflection, bind handling, and result handling.
- Map reflected Athena ARRAY columns to the appropriate PyAthena SQLAlchemy ARRAY type.
- Open the SQLAlchemy `array_type` requirement when the compliance tests pass.
- Preserve compatibility for existing AthenaArray users.

## Acceptance criteria

- SQLAlchemy ARRAY compliance tests pass against Athena engine v3.
- Reflected ARRAY element types are preserved.
- Existing public ARRAY behavior remains backward compatible or the compatibility impact is documented.
- Compile-only regression tests cover type rendering where possible.
- `just lint` passes.
- Athena tests are limited to the affected nodes and run with `-n 1`.

PR #774, opened 2026-09-19
Support native SQLAlchemy ARRAY types and typed round trips

laughingman7743:
## WHAT

Support native SQLAlchemy ARRAY I/O through `AthenaArray`: preserve reflected element types, render bound/literal arrays, and return typed Python values for nested ARRAY/MAP/ROW values, NULLs, decimal, binary and temporal elements.
Preserve element and outer TypeDecorator processors and existing DB API IN-list formatting.

Keep DISTINCT, ordering and compound SELECT operations on native values, then project outer results through JSON transport.
Resolve ORDER BY labels against selected and FROM columns, reject unsupported projection/order references, and warn/fall back for unrecognized nested reflection types.
Require explicit Numeric precision for implicit decimal ARRAY binds to prevent silent rounding while preserving explicit CAST semantics.

Group the ARRAY type, type inspection, value processing and JSON projection in `array.py`, with tests organized by the same responsibilities.
Place MAP, STRUCT and temporal dependencies in focused modules and retain existing public `types.py` imports.

The ARRAY documentation distinguishes typed SQLAlchemy SELECT results from existing direct REST cursor/untyped SQL conversion.
It gives concrete unsupported and ambiguous native-text examples, explains explicit JSON conversion, and removes the misleading raw-string migration example.

Compatibility and limits:

- Reflected arrays use AthenaArray instead of String; nested DDL uses INT/STRUCT.
- Raw SQL and unknown-element arrays retain the documented cursor fallback. Rewritten ARRAY queries require explicit SELECT columns; literal SQL expressions require labels.
- Textual ORDER BY supports selected names/lists and ordinals. Use SQLAlchemy expressions for other expressions.
- FLOAT/REAL uses 32 bits; use Double for 64-bit values. Supported decimal binds declare precision and scale.

Validation for `50f07479bf2eb4e0a8fb9cbad8274c263e1be671`:

- All 18 current-head CI checks pass: ordinary, synchronous SQLAlchemy and asynchronous SQLAlchemy tests on Python 3.10–3.14, documentation lint/build, and benchmark tooling. [Test run](https://github.com/pyathena-dev/PyAthena/actions/runs/35553777307).
- `just format`, `just lint`, Markdown lint and the current-source HTML build pass for the documentation repair. Converter examples and typed SELECT compilation were checked locally without AWS queries. The unchanged implementation previously passed 164 focused type/compiler cases.
- SQLAlchemy 2.0.0 public import/compile/result smoke passes; the main suite uses locked 2.0.46. The built wheel includes the new modules, and a nested type pickle from the previous module resolves through compatibility exports.
- Both self-review perspectives and independent Claude reviews are recorded inline, including repairs and two low-severity diagnostic limits for unsupported SQL forms. The class, module, test-grouping and final documentation follow-ups are CLEAN.

## WHY

ARRAY reflection returned String, generic ARRAY lost element types, and list parameters used IN-list syntax.
This enables native typed ARRAY I/O and the SQLAlchemy array_type compliance requirement on Athena engine v3.

Refs #755. Stack: #774 → #775 → #776.
Expression operators and partial updates follow in their own PRs.


pyathena/sqlalchemy/types.py
```diff
@@ -256,21 +273,277 @@ class AthenaArray(TypeEngine[list[Any]]):
...
 
-    @property
-    def python_type(self) -> type:
-        return list
+    def bind_expression(self, bindvalue):
+        """Cast a bound ARRAY value to its declared Athena element type."""
+        # The cast also gives empty arrays and NULL-only arrays their element type.
+        return cast(bindvalue, self)
+
+    def bind_processor(self, dialect):
```
laughingman7743: Self-review round one — implementation and regression coverage: CLEAN within `47024e2084ce5c9a338b190d1934151e22c19817..8b0fb91c2a469b980aef1d4f02ce52f575d768cf` (all 11 changed files).

Checked ARRAY adaptation and recursive reflection, Hive DDL versus DML casts, bound/literal formatting, recursive typed decoding, NULL/empty collections, decimal and binary preservation, nested fields, Enum/TypeDecorator processors, outer-result transport, native DISTINCT/order/UNION semantics, and the existing integration fixtures. Traced the shared compiler/type paths through synchronous and asynchronous dialects and the five cursor adapters. Compared the six-commit series with the pre-rebase range and checked relevant upstream changes: the merged Binary/CTE tests, AthenaBinary colspecs, metadata cache/error handling, and dataframe conversion fixes remain intact. Conflict repairs are additive imports/colspecs/test declarations, plus import ordering.

Validation so far: `just format`, `just lint` (mypy: 86 source files), and 195 type/compiler/formatter tests pass. The existing full synchronous SQLAlchemy recipe also passes: 519 passed, 766 skipped (775.17 s). Async ARRAY tests (9) and all five cursor cases also pass. CI remains pending; this static CLEAN result does not assert their completion.

The final maintenance-only rebase preserves all six patches (range-diff `=` throughout). PyAthena sources, tests and the changed ARRAY documentation are byte-identical to the locally tested revision `15611023aa40cccdd852c296e90112f76e93b4a6`; format/lint/mypy/cfn-lint were rerun successfully on this published revision. No historical test result is presented as a separate rerun.

docs/sqlalchemy.md
```diff
@@ -1003,7 +1003,52 @@ value = map_data['key1']  # Direct access
...
+Use `dimensions=N` for a fixed number of dimensions with the standard SQLAlchemy type.
+Without it, the table column has one dimension.
+The existing `AthenaArray(AthenaArray(Integer))` spelling also remains supported; do not combine nested ARRAY types with `dimensions`.
+`AthenaArray()` still defaults to string elements, and the uppercase `pyathena.sqlalchemy.types.ARRAY` alias remains available.
+Set `as_tuple=True` to return tuples at each array dimension instead of lists.
+
+Bound parameters, multiple parameter sets, and SQLAlchemy literals use Athena's `ARRAY[...]` constructors.
+An empty list represents an empty array, `None` represents SQL NULL, and individual elements can also be NULL.
+Ordinary DB API list and tuple parameters retain their existing `IN (...)` formatting.
+
```
laughingman7743, line 1042: Self-review round two — caller compatibility, claims and operational effects: CLEAN within `47024e2084ce5c9a338b190d1934151e22c19817..8b0fb91c2a469b980aef1d4f02ce52f575d768cf` (all 11 changed files).

Audited the documentation and API claims against constructors, reflected type trees, dialect result processing and SQLAlchemy dependency sources. Existing AthenaArray defaults/nested construction and ordinary DB API IN parameters remain supported. Typed outer SELECT results use a JSON envelope; predicates and intermediate subqueries retain native ARRAY values. Unknown-element and raw-text paths retain the documented cursor fallback. Checked ordering restrictions, decimal/string-length behavior, binary/date decoding, decorator processing, resource cleanup in the cross-cursor test, and shared sync/async reflection. The change adds no metadata retry policy or cache ownership changes; no throttling/performance claim is made. Current dependency validation uses the locked SQLAlchemy 2.0.46; no claim of exhaustive historical dependency-version coverage.

Current documentation lint and HTML generation pass (154 Sphinx warnings, not a warning-free build). Local unit/compile/formatter results are current; full runtime/CI evidence will be recorded separately. Reflected ARRAY types and nested compiled SQL intentionally differ from prior releases as documented.

The final maintenance-only rebase preserves all six patches (range-diff `=` throughout). PyAthena sources, tests and the changed ARRAY documentation are byte-identical to the locally tested revision `15611023aa40cccdd852c296e90112f76e93b4a6`; format/lint/mypy/cfn-lint were rerun successfully on this published revision. No historical test result is presented as a separate rerun.

pyathena/sqlalchemy/compiler.py
```diff
@@ -228,6 +256,133 @@ class AthenaStatementCompiler(SQLCompiler):
...
+            rendered = self.process(self._array_result_select(cs), **kw)
+            self._result_columns = [
+                entry._replace(objects=(*entry.objects, original))
+                for entry, original in zip(self._result_columns, original_columns, strict=True)
+            ]
+            return rendered
+        return super().visit_compound_select(cs, asfrom=asfrom, compound_index=compound_index, **kw)
+
+    def _array_result_select(self, statement):
+        columns = list(statement.selected_columns)
```
laughingman7743, line 315: Additional self-review — FINDINGS. `select(text("id"), table.c.array_value).order_by(table.c.array_value)` keeps `id` in the inner query but silently omits it from the outer result because `selected_columns` excludes TextClause projections. A literal `*` also lets the rewrite expose duplicate/helper columns. Reproduced by compiling against this published revision. The rewrite must preserve the result shape or reject projections whose columns cannot be represented safely, with an explicit SQLAlchemy-expression alternative and regression coverage.

pyathena/sqlalchemy/types.py
```diff
@@ -256,21 +273,277 @@ class AthenaArray(TypeEngine[list[Any]]):
...
+        """Return a processor that renders typed Athena array literals."""
+
+        def process(value):
+            return _literal_complex(value, self, dialect)
+
+        return process
+
+    def column_expression(self, colexpr):
+        """Serialize an outer result column without changing its native SQL type."""
+        return colexpr if _has_unknown_array_element(self) else _ArrayResult(colexpr, self)
```
laughingman7743: Additional self-review — FINDINGS. An outer TypeDecorator with `impl = AthenaArray(Integer)` and a custom `process_result_value` loses that processor: the compiled result entry is AthenaArray rather than the declared decorator. Ordering that decorated ARRAY by a selected label also bypasses the native-array wrapper, so `ORDER BY label` can compare the serialized projection. Reproduced by inspecting the compiled SQL and result type. Preserve the declared result type and recognize decorated ARRAY results when deciding whether to rewrite ordering/compound SELECTs.

pyathena/sqlalchemy/compiler.py
```diff
@@ -228,6 +256,133 @@ class AthenaStatementCompiler(SQLCompiler):
...
+        inner = statement.order_by(None).limit(None).offset(None)
+        ordering = []
+        hidden: list[Any] = []
+
+        def resolve_label(element: Any, **kw: Any) -> Any:
+            if isinstance(element, _textual_label_reference):
+                return statement.selected_columns[element.element]
```
laughingman7743: Relayed independent static review by Claude Code (`claude-fable-5-1`, Max profile, effort high), session `aaed43e4-c7fe-4991-b6c0-1115f4085080`, on `47024e2084ce5c9a338b190d1934151e22c19817..8b0fb91c2a469b980aef1d4f02ce52f575d768cf`: FINDINGS. The actual result was collected; all 11 changed files, their callers/tests/conventions, and relevant SQLAlchemy 2.0.46 sources were covered. Read-only tools only; no reviewer execution or network access. Source snapshot and worktree remained unchanged. The confirmed string-label regression is `select(t.c.items).order_by("id")` when id is only in the FROM table: this lookup raises bare KeyError. Table-qualified SQLAlchemy labels are also affected. Native SQLAlchemy resolves these against selected labels and FROM columns. Resolve them using that contract and produce CompileError for genuinely unknown names. Author reproduction/repair is in progress.

pyathena/sqlalchemy/compiler.py
```diff
@@ -307,6 +467,72 @@ def visit_cast(self, cast: Cast[Any], **kwargs):
...
+            return f"ROW({fields})"
+        if isinstance(type_, types.String):
+            return "VARCHAR"
+        if isinstance(type_, (types.LargeBinary, types.BINARY, types.VARBINARY)):
+            return "VARBINARY"
+        if isinstance(type_, getattr(types, "Double", get_double_type())):
+            return "DOUBLE"
+        if isinstance(type_, types.Float):
+            return "REAL"
+        return self.dialect.type_compiler_instance.process(type_)
```
laughingman7743, line 547: Independent review: the unconditional ARRAY bind CAST renders unparameterized Numeric as bare DECIMAL and Float as REAL. The reviewer confirmed rendering, but did not execute Athena; it warned of decimal rounding and double-to-real narrowing when the model is less specific than an existing column. Decimal values with explicit precision/scale are already covered. The author is checking the runtime effect and distinguishing unsupported/unbounded declarations from intentional FLOAT/REAL semantics.

pyathena/sqlalchemy/base.py
```diff
@@ -540,6 +544,36 @@ def _get_column_type(self, type_: str):
...
+                    else r'\s*("(?:[^"]|"")*"|`(?:[^`]|``)*`|[^\s:]+)(?:\s*:\s*|\s+)(.+)'
+                )
+                match = re.fullmatch(
+                    pattern,
+                    field,
+                )
+                if match is None:
+                    raise ValueError(f"Invalid ROW field: {field!r}")
+                field_name, field_type = match.groups()
+                field_name = field_name.strip()
```
laughingman7743, line 574: Independent review: nested ARRAY reflection can raise ValueError for malformed MAP/ROW signatures instead of warning and degrading as unknown types do elsewhere. The parse failure is confirmed statically; whether Athena returns the proposed anonymous-row spelling for a reflected relation is unconfirmed. Preserve reflection robustness without translating service/API failures into missing columns.

pyathena/sqlalchemy/types.py
```diff
@@ -233,13 +247,16 @@ class MAP(AthenaMap):
     __visit_name__ = "MAP"
 
 
-class AthenaArray(TypeEngine[list[Any]]):
+class AthenaArray(sqltypes.ARRAY[Any]):
```
laughingman7743: Independent review also reported lower-confidence/limited-scope items: Uuid/Time/Interval leaves are not decoded, timezone-bearing timestamp text may be rejected by fromisoformat, early SQLAlchemy versions may not support runtime ARRAY[Any] subclassing, a fully custom Formatter mapping lacks the private complex-parameter handler, and multidimensional decoding constructs child ARRAY types per element. These are recorded separately from confirmed regressions; no measured performance regression or live support for those Athena types was established. The SQLAlchemy 2.0.0 official source has now been inspected for the generic-base compatibility concern. Dispositions and any repairs will follow in these threads. The documented textual ORDER BY restriction was identified as a compatibility limit, not a new finding.

pyathena/sqlalchemy/types.py
laughingman7743: Self-review round one, repair follow-up: CLEAN on `47024e2084ce5c9a338b190d1934151e22c19817..f472cc2cb5f7fbaaa60e5d1966f3db7692f60598`. Verified both previous objects and range-diff: the original six commits are unchanged; one six-file repair is added. Reviewed the full expanded result-type/compiler/test/documentation contract, including outer TypeDecorator and dialect variants, plus the unchanged surfaces from the initial eleven-file inventory.

The repair preserves declared result processors while using the adapted ARRAY type for transport, resolves string labels using SQLAlchemy's selected/FROM-column rules, rejects untracked text/wildcard projections before they can lose or add result columns, warns and degrades unknown nested signatures without masking service errors, and rejects precision-free Numeric only in implicit bind casts. Explicit user CASTs remain valid. Child ARRAY types are reused within each collection; no measured speedup is claimed.

Validation: format/lint/mypy/cfn-lint and 153 type/compiler tests pass. The full synchronous suite passed (520 passed, 766 skipped); the final implicit-cast refinement then passed all ten synchronous ARRAY tests, all ten asynchronous ARRAY tests, and all five cursor cases including supported unload modes. Current CI and the independent follow-up remain pending.

docs/sqlalchemy.md
laughingman7743: Self-review round two, repair follow-up: CLEAN on `47024e2084ce5c9a338b190d1934151e22c19817..f472cc2cb5f7fbaaa60e5d1966f3db7692f60598`. Audited the new restrictions and expanded outer-decorator contract against callers, dependency source and runtime evidence. The untracked-projection error names explicit-column alternatives; normal string labels retain SQLAlchemy resolution, and raw text queries that do not receive typed transport retain their existing path. Decimal precision is required only for implicit bind conversion, preserving intentionally explicit casts. A real SELECT reproduced unbounded Numeric 1.50 becoming 2; FLOAT/REAL rounding is the existing declared type mapping and is now distinguished from Double in the docs. Removed a legacy raw-ROW example's unsupported numeric-result assertion.

Reflection fallback is limited to parsing failures; API errors/cache ownership are unchanged. SQLAlchemy 2.0.0 actually accepts ARRAY[Any], and an isolated import/compile smoke covers labels and both cast paths, so the suspected generic-base defect was not retained. UUID/TIME/INTERVAL/timezone support and arbitrary replacement Formatter mappings remain outside the claimed Athena-table-type/default-formatter support. Allocation simplification has no claimed benchmark gain. Markdown lint passes; HTML generation succeeded with 154 warnings. Runtime results and the final narrow refinement are distinguished above; pending CI/reviewer work is not reported as passed.

pyathena/sqlalchemy/compiler.py
laughingman7743: Fixed in `f472cc2cb5f7fbaaa60e5d1966f3db7692f60598`. Rewritten ARRAY queries reject text()/wildcard SELECT projections before they can silently omit columns or expose helper columns. Explicit column expressions/literal_column() remain supported. Parametrized compile tests cover ORDER BY, DISTINCT and UNION, and the docs state the alternatives.

pyathena/sqlalchemy/types.py
laughingman7743: Fixed in `f472cc2cb5f7fbaaa60e5d1966f3db7692f60598`. The result expression retains the declared type (including outer TypeDecorator/variants), with a separate adapted ARRAY type for transport. Rewrite detection resolves the dialect implementation. Unit tests and both live sync/async ARRAY suites verify native ordering and the custom tuple result processor.

pyathena/sqlalchemy/compiler.py
laughingman7743: Fixed in `f472cc2cb5f7fbaaa60e5d1966f3db7692f60598`. SELECT label resolution uses SQLAlchemy's selected/FROM label map; compound SELECTs retain their selected-column name scope. Nonselected FROM columns become hidden ordering columns, qualified selected labels work, and unresolved names produce CompileError. Both live ARRAY suites verify the new cases.

pyathena/sqlalchemy/compiler.py
laughingman7743: Addressed in `f472cc2cb5f7fbaaa60e5d1966f3db7692f60598`. A real Athena SELECT reproduced Numeric() `[Decimal("1.50")]` becoming `[Decimal("2")]`. Implicit ARRAY bind casts now require Numeric precision, including nested fields, instead of silently rounding. Explicit user CASTs are preserved and verified live in sync/async tests. Float narrowing was also observed, but it matches the dialect's existing FLOAT/REAL (32-bit) declaration; the docs direct 64-bit callers to Double. No default floating-point type mapping was changed.

pyathena/sqlalchemy/base.py
laughingman7743: Fixed in `f472cc2cb5f7fbaaa60e5d1966f3db7692f60598`. Unrecognized nested ARRAY signatures warn and fall back to NullType; malformed MAP/ROW/numeric signatures are covered by standalone parser/reflection tests. Only local type parsing errors are caught. Metadata API failures and cache/error policies remain unchanged; the proposed real-catalog anonymous-row trigger was not claimed as observed.

pyathena/sqlalchemy/types.py
laughingman7743: Disposition on `f472cc2cb5f7fbaaa60e5d1966f3db7692f60598`: SQLAlchemy 2.0.0 accepts ARRAY[Any] at runtime, and an isolated smoke verifies imports, label resolution and both decimal-cast paths. The suspected generic-base incompatibility was not reproduced, so no generic-base workaround was retained. UUID/TIME/INTERVAL/timezone-aware table types remain outside this dialect's claimed supported element set; arbitrary Formatter implementations replacing the default handlers are not covered by default-formatter behavior. The multidimensional item type is now computed once per collection rather than once per element; no measured performance gain is claimed.

pyathena/sqlalchemy/compiler.py
```diff
@@ -228,6 +257,153 @@ class AthenaStatementCompiler(SQLCompiler):
...
+        return select_stmt
+
+    def visit_compound_select(self, cs, asfrom=False, compound_index=None, **kw):
+        if (
+            not self.stack
+            and not asfrom
+            and any(self._has_array_result(column) for column in cs.selected_columns)
+        ):
+            original_columns = list(cs.selected_columns)
+            rendered = self.process(self._array_result_select(cs), **kw)
```
laughingman7743, line 281: Relayed independent static review: Claude claude-fable-5-1, verified first-party Max profile max, effort high; session c0dddefd-b833-42bd-ae76-b2245bf4e60f. Full diff 47024e2084ce5c9a338b190d1934151e22c19817..f472cc2cb5f7fbaaa60e5d1966f3db7692f60598 plus changed source/tests, direct callers and SQLAlchemy 2.0.46 internals. Read/Glob/Grep only; no commands, tests, network, edits or PR discussion. Snapshot and worktree unchanged. Verdict FINDINGS.

Confirmed by author compilation: an ordered top-level UNION is rewritten twice. Ordering by an unselected original-table column can add that table to an intermediate FROM clause and multiply rows. DISTINCT leaves that column out of scope. Repair will guard rewrite re-entry and reject unadaptable ordering expressions.

pyathena/sqlalchemy/compiler.py
```diff
@@ -228,6 +257,153 @@ class AthenaStatementCompiler(SQLCompiler):
...
+            isinstance(column, TextClause)
+            or (getattr(column, "is_literal", False) and column.name.rstrip().endswith("*"))
+            for column in statement._all_selected_columns
+        ):
+            raise exc.CompileError(
+                "Ordered, DISTINCT, and compound ARRAY results require explicit SELECT columns; "
+                "use SQLAlchemy column expressions or literal_column() instead of text(), "
+                "and select(table) instead of a wildcard"
+            )
+        columns = list(statement.selected_columns)
```
laughingman7743, line 315: Relayed independent static review: Claude claude-fable-5-1, verified first-party Max profile max, effort high; session c0dddefd-b833-42bd-ae76-b2245bf4e60f. Full diff 47024e2084ce5c9a338b190d1934151e22c19817..f472cc2cb5f7fbaaa60e5d1966f3db7692f60598 plus changed source/tests, direct callers and SQLAlchemy 2.0.46 internals. Read/Glob/Grep only; no commands, tests, network, edits or PR discussion. Snapshot and worktree unchanged. Verdict FINDINGS.

Confirmed by author compilation: an unlabeled literal_column("cardinality(items)") survives as a literal expression inside the subquery, but the outer query references a nonexistent output name. Require an explicit label for literal expressions; keep plain literal identifiers supported and document the constraint.

pyathena/sqlalchemy/types.py
```diff
@@ -256,21 +273,279 @@ class AthenaArray(TypeEngine[list[Any]]):
...
+    if isinstance(type_, types.JSON):
+        return value
+    if isinstance(type_, types.Boolean):
+        return value if isinstance(value, bool) else value.lower() == "true"
+    if isinstance(type_, types.Integer):
+        return int(value)
+    if isinstance(type_, types.Numeric):
+        return Decimal(value) if type_.asdecimal else float(value)
+    if isinstance(type_, types.DateTime):
+        return value if isinstance(value, datetime) else datetime.fromisoformat(value)
```
laughingman7743: Relayed independent static review: Claude claude-fable-5-1, verified first-party Max profile max, effort high; session c0dddefd-b833-42bd-ae76-b2245bf4e60f. Full diff 47024e2084ce5c9a338b190d1934151e22c19817..f472cc2cb5f7fbaaa60e5d1966f3db7692f60598 plus changed source/tests, direct callers and SQLAlchemy 2.0.46 internals. Read/Glob/Grep only; no commands, tests, network, edits or PR discussion. Snapshot and worktree unchanged. Verdict FINDINGS.

Confirmed decoding defect: AthenaArray(AthenaDate()) and AthenaArray(AthenaTimestamp()) return strings. The reviewer said binding works, but the author additionally reproduced an AttributeError compiling their implicit ARRAY cast because these helper classes lack type compiler dispatch. Repair will cover type compilation and bound/literal result round trips.

pyathena/sqlalchemy/types.py
```diff
@@ -233,13 +247,16 @@ class MAP(AthenaMap):
     __visit_name__ = "MAP"
 
 
-class AthenaArray(TypeEngine[list[Any]]):
+class AthenaArray(sqltypes.ARRAY[Any]):
     """SQLAlchemy type for Athena ARRAY complex type.
 
     ARRAY represents an ordered collection of elements of the same type.
```
laughingman7743: Relayed independent static review: Claude claude-fable-5-1, verified first-party Max profile max, effort high; session c0dddefd-b833-42bd-ae76-b2245bf4e60f. Full diff 47024e2084ce5c9a338b190d1934151e22c19817..f472cc2cb5f7fbaaa60e5d1966f3db7692f60598 plus changed source/tests, direct callers and SQLAlchemy 2.0.46 internals. Read/Glob/Grep only; no commands, tests, network, edits or PR discussion. Snapshot and worktree unchanged. Verdict FINDINGS.

Disposition of lower-confidence/nit items: SQLAlchemy 2.0.0 actually supports ARRAY[Any] through inherited __class_getitem__; an isolated import/compiler smoke and primary source confirm this, so no compatibility workaround is warranted for that hypothesized failure. Inner as_tuple does not override the documented outer ARRAY collection-shape setting; retain that contract. No support for arbitrary untested element types is claimed. Textual ORDER BY limitations and raw-result fallback remain documented.

pyathena/sqlalchemy/types.py
laughingman7743: Self-review round 1, additional repair: CLEAN. Full series base 47024e2084ce5c9a338b190d1934151e22c19817, head 2b03ac8be983430f454881ca1d28edbf91ec5556; repair comparison f472cc2cb5f7fbaaa60e5d1966f3db7692f60598..2b03ac8be983430f454881ca1d28edbf91ec5556. This is author review, not independent review.

- CompoundSelect re-entry is prevented on the generated node only; inner native ARRAY types and result-column remapping remain intact.
- Unselected ordering columns in DISTINCT/UNION fail before SQL execution; ordinary SELECT helper ordering and expressions derived from selected columns remain supported.
- Literal-expression labeling is explicit and guards avoid dropping selected output; simple identifiers stay supported.
- AthenaDate/AthenaTimestamp dispatch matches existing Athena type compiler methods; result conversion, bound values, literals and nested ARRAY recursion use the declared Python date types.
- Standalone parametrized compiler/type tests reproduce original failures; integration cases extend existing NativeArrayTest and real connection fixtures.

Validation: required format/lint and 164 pure type/compiler tests pass; SQLAlchemy 2.0.0 isolated compiler smoke passes. Final synchronous native ARRAY tests pass (11); asynchronous validation and current CI are recorded separately. The earlier full local SQLA run is not represented as a run on this revision.

docs/sqlalchemy.md
laughingman7743: Self-review round 2, additional repair: CLEAN. Full series base 47024e2084ce5c9a338b190d1934151e22c19817, head 2b03ac8be983430f454881ca1d28edbf91ec5556; repair comparison f472cc2cb5f7fbaaa60e5d1966f3db7692f60598..2b03ac8be983430f454881ca1d28edbf91ec5556. This is author review, not independent review.

- UNION output is wrapped once, tested by one FROM subquery assertion and real native ordered results.
- DISTINCT/compound ordering cannot use original-table columns absent from the derived source; documentation narrows FROM label resolution accordingly.
- Documentation now requires labels for nonidentifier literal SQL expressions; no claim of general SQL text parsing.
- Direct temporal helper types use the same Athena DATE/TIMESTAMP SQL names and are explicitly tested for Python return types.
- No retry, cache, AWS API, concurrency or fixture lifecycle changes; existing metadata errors remain visible.
- Older ARRAY[Any] hypothesis is refuted for actual 2.0.0; no full historical version matrix or full final-head suite claim.

Validation: required format/lint and 164 pure type/compiler tests pass; SQLAlchemy 2.0.0 isolated compiler smoke passes. Final synchronous native ARRAY tests pass (11); asynchronous validation and current CI are recorded separately. The earlier full local SQLA run is not represented as a run on this revision.

pyathena/sqlalchemy/compiler.py
laughingman7743: Relayed independent repair follow-up: Claude claude-fable-5-1, verified first-party Max profile max, effort high; session 205ba42e-feaa-451c-a56a-fe4cf5312bf0. Base 47024e2084ce5c9a338b190d1934151e22c19817; previous reviewed head f472cc2cb5f7fbaaa60e5d1966f3db7692f60598; current head 2b03ac8be983430f454881ca1d28edbf91ec5556. Read-only repair diff/series comparison plus full affected temporal, compiler, ordering, result-map and dependency contracts; no commands, tests, network or PR discussion. Actual result collected; snapshot and worktree unchanged. Verdict FINDINGS (two low-severity diagnostic gaps; reviewer found no regression in the repaired contracts).

The rewrite annotation and FROM adaptation check fix the original double wrapping, cartesian-product and out-of-scope direct-column cases. The reviewer also found that _from_objects does not inspect a scalar subquery: an ORDER BY scalar subquery that refers to an unselected original-table column can stop correlating after DISTINCT. Author compilation confirms that shape. Deferred as a diagnostic limit for this unsupported query: DISTINCT ordering must refer to selected results, and this guard is not a general SQL/subquery validity checker. No support for arbitrary correlated ORDER BY scalar subqueries is claimed. Existing supported ordering forms and result mapping passed review and the live ARRAY tests.

pyathena/sqlalchemy/compiler.py
laughingman7743: Relayed independent repair follow-up: Claude claude-fable-5-1, verified first-party Max profile max, effort high; session 205ba42e-feaa-451c-a56a-fe4cf5312bf0. Base 47024e2084ce5c9a338b190d1934151e22c19817; previous reviewed head f472cc2cb5f7fbaaa60e5d1966f3db7692f60598; current head 2b03ac8be983430f454881ca1d28edbf91ec5556. Read-only repair diff/series comparison plus full affected temporal, compiler, ordering, result-map and dependency contracts; no commands, tests, network or PR discussion. Actual result collected; snapshot and worktree unchanged. Verdict FINDINGS (two low-severity diagnostic gaps; reviewer found no regression in the repaired contracts).

The explicit-label repair passed review. The reviewer noted that bare keyword expressions such as literal_column("current_date") match the plain-identifier guard. Author compilation confirms that they can still reach Athena with an invalid outer reference. The documented contract already requires labels for literal SQL expressions; literal_column("current_date").label("today") compiles correctly. Retain this documented requirement rather than claim complete SQL lexical validation. This is a remaining diagnostic gap for unsupported unlabeled input, not a supported-use regression.

pyathena/sqlalchemy/types.py
laughingman7743: Relayed independent repair follow-up: Claude claude-fable-5-1, verified first-party Max profile max, effort high; session 205ba42e-feaa-451c-a56a-fe4cf5312bf0. Base 47024e2084ce5c9a338b190d1934151e22c19817; previous reviewed head f472cc2cb5f7fbaaa60e5d1966f3db7692f60598; current head 2b03ac8be983430f454881ca1d28edbf91ec5556. Read-only repair diff/series comparison plus full affected temporal, compiler, ordering, result-map and dependency contracts; no commands, tests, network or PR discussion. Actual result collected; snapshot and worktree unchanged. Verdict FINDINGS (two low-severity diagnostic gaps; reviewer found no regression in the repaired contracts).

Temporal type dispatch, nested DDL/CAST, bind/literal formatting and typed result conversion passed the follow-up. Author validation on the final head: 164 pure type/compiler tests, 11 synchronous ARRAY cases and 11 asynchronous ARRAY cases pass; the new temporal test covers both bound and literal_execute paths. The reviewer separately observed a scalar Date/Timestamp literal_execute render_bind_cast limitation outside this repair. The render flags already exist in merge base 47024e2; it is not introduced by this ARRAY change and is outside this PR.

pyathena/sqlalchemy/types.py
laughingman7743: Self-review round one: CLEAN. Series base 47024e2084ce5c9a338b190d1934151e22c19817, head f8511079cc6df9ceb4c39c0b24187e925760d359; bounded refactor comparison 2b03ac8be983430f454881ca1d28edbf91ec5556..f8511079cc6df9ceb4c39c0b24187e925760d359 plus all affected type/compiler callers. 

Reviewed each moved branch and processor call site. AthenaArray retains the SQLAlchemy ARRAY contract; _ArrayValueProcessor owns the array/dialect context and all recursive parameter, literal and result conversions. _ArrayTypeInspector owns dimension traversal, dialect-specific TypeDecorator resolution and unknown-element detection shared with the compiler. No value is retained between processor calls. JSON fallback, NULL handling, tuple propagation (including map-key behavior), custom processor overrides and variants remain unchanged. The renamed _ArrayJSONProjection retains both original result type and native array type in cache traversal, and its compiler visit name matches.

Required format/lint and the unchanged 164 type/compiler tests pass; SQLAlchemy 2.0.0 import, decorated UNION result, nested tuple and literal smoke pass. Existing behavioral tests are reused; no implementation-mirroring tests were added.

docs/sqlalchemy.md
laughingman7743: Self-review round two: CLEAN. Series base 47024e2084ce5c9a338b190d1934151e22c19817, head f8511079cc6df9ceb4c39c0b24187e925760d359; bounded refactor comparison 2b03ac8be983430f454881ca1d28edbf91ec5556..f8511079cc6df9ceb4c39c0b24187e925760d359 plus all affected type/compiler callers. 

Audited the class docstrings and the claim that this is a responsibility refactor. The public constructor, ARRAY alias, reflection result types, SQL output and conversion rules remain unchanged. ColumnElement models the JSON projection expression; it is not a row-result container. Keeping its original type preserves outer TypeDecorator processing. The per-compiler memoized inspector retains only the dialect, not metadata or query results; no AWS request, retry or cache freshness behavior changes.

Compared SQLAlchemy PostgreSQL ARRAY (2.0.46) and Trino datatype.py at c4b2c5781c82db8e0f7ddedbd51b2d480ff6ef9b. PostgreSQL exposes processors through ARRAY and shares dimension traversal; Trino separates TypeEngine/TypeDecorator roles from formatting helpers. PyAthena keeps those type/SQL-expression roles while its value processor handles Athena JSON transport and nested MAP/ROW data. The generic ARRAY _apply_item_processor does not itself cover heterogeneous ROW fields or MAP keys/values. No new backend dependency or performance improvement is claimed. Current-head CI and independent refactor review remain pending.

pyathena/sqlalchemy/types.py
```diff
@@ -256,21 +277,310 @@ class AthenaArray(TypeEngine[list[Any]]):
...
+                type_.key_type
+            ) or _ArrayTypeInspector.has_unknown_element(type_.value_type)
+        if isinstance(type_, AthenaStruct):
+            return any(
+                _ArrayTypeInspector.has_unknown_element(field) for field in type_.fields.values()
+            )
+        return isinstance(type_, types.NullType)
+
+
+class _ArrayValueProcessor:
```
laughingman7743: Relayed independent refactor review: CLEAN. Claude claude-fable-5-1, verified first-party Max profile max, effort high; session 6a64cec8-ec06-40ff-b87e-2bd698f4339d. Series base 47024e2084ce5c9a338b190d1934151e22c19817, current head f8511079cc6df9ceb4c39c0b24187e925760d359; literal repair and series comparison from 2b03ac8be983430f454881ca1d28edbf91ec5556. Read/Glob/Grep only; no commands, tests, network, edits or PR discussion. Actual result collected; review snapshot and PR worktree unchanged.

The reviewer compared every moved helper with its new method and found no semantic change. Coverage: type/dialect resolution, bind/literal/result branch order and custom processors, map-key tuple handling, JSON fallback/envelope, references retained by bound methods and SQLAlchemy type-memo caching, compiler inspector initialization, projection visit dispatch, cache traversal and original result-type retention, all direct callers and existing tests/docs. The classes have distinct responsibilities and the processor carries no per-call value state.

Limits: this is static review. Existing behavioral tests provide indirect coverage of the new classes; no tests of implementation structure are required. The reviewer could not prove from the history-free snapshot that the old visit method was unreleased; the author verified it is absent from merge base 47024e2084ce5c9a338b190d1934151e22c19817. The visit rename therefore replaces an internal name introduced by this unmerged series. The reviewer explicitly left unchanged type-inspection behavior outside this refactor closed. Final-head CI is still running.

pyathena/sqlalchemy/types.py
laughingman7743: Self-review round one: CLEAN. Series base 47024e2084ce5c9a338b190d1934151e22c19817, current head ad5e526cd7fbcc504b4a2667a449997a1deb2958; module-split comparison f8511079cc6df9ceb4c39c0b24187e925760d359..ad5e526cd7fbcc504b4a2667a449997a1deb2958, plus all affected import/compiler/package/docs/test contracts.

ARRAY and its inspector, value processor and JSON projection now live together in array.py. MAP, STRUCT and temporal types are independent leaf modules; array.py imports those leaves, and types.py re-exports the established public type names while retaining the smaller scalar helpers. The compiler imports its private ARRAY inspector from array.py. This dependency graph has no reverse import from a leaf to types.py or compiler.py. Public aliases resolve to the same type classes.

AST comparisons confirmed all 15 moved source class/function definitions and all 48 test definitions/decorators are unchanged. The same 164 tests pass after relocation, including existing fixtures and parametrization. Required format/lint passes (90 source files). SQLAlchemy 2.0.0 public import/compile/result smoke passes. A nested type pickle created with the previous types module resolves through the compatibility exports.

docs/sqlalchemy.md
laughingman7743: Self-review round two: CLEAN. Series base 47024e2084ce5c9a338b190d1934151e22c19817, current head ad5e526cd7fbcc504b4a2667a449997a1deb2958; module-split comparison f8511079cc6df9ceb4c39c0b24187e925760d359..ad5e526cd7fbcc504b4a2667a449997a1deb2958, plus all affected import/compiler/package/docs/test contracts.

Audited the claims that related implementations are grouped and public type import paths remain available. Existing production callers and tests continue using types.py exports; only the private compiler helper import changes. The documentation still addresses the established types.AthenaArray/Map/Struct/Date/Timestamp paths. The API HTML build succeeds (154 existing Sphinx warnings). A locally built wheel includes array.py, map.py, struct.py, temporal.py and types.py. The tests were moved to files matching the source type modules without changing their AST or adding tests of implementation structure.

The layout follows the official PostgreSQL dialect pattern of type-specific modules plus shared re-exports. No new optional dependencies, public conversion behavior, AWS requests/retries or metadata cache behavior are introduced. Import smoke, the specific previous-module pickle check and wheel inspection are bounded evidence, not a guarantee about every historical serialized object. Independent module-split review and final-head CI remain pending.

pyathena/sqlalchemy/array.py
```diff
@@ -0,0 +1,355 @@
...
+from sqlalchemy.sql.visitors import InternalTraversal
+
+from pyathena.formatter import _ComplexParameter
+from pyathena.sqlalchemy.map import AthenaMap
+from pyathena.sqlalchemy.struct import AthenaStruct
+from pyathena.sqlalchemy.temporal import AthenaDate, AthenaTimestamp
+
+
+class AthenaArray(sqltypes.ARRAY[Any]):
+    """SQLAlchemy type for Athena ARRAY complex type.
```
laughingman7743, line 24: Relayed independent module-split review: CLEAN. Claude claude-fable-5-1, verified first-party Max profile max, effort high; session c5a36c59-e333-4301-b26e-a0c25ea3f2e1. Series base 47024e2084ce5c9a338b190d1934151e22c19817; previous head f8511079cc6df9ceb4c39c0b24187e925760d359; current head ad5e526cd7fbcc504b4a2667a449997a1deb2958. Literal repair and series comparison, all changed files and affected callers/contracts. Read/Glob/Grep only; no commands, tests, network, edits or PR discussion. Actual result collected; snapshot and PR worktree unchanged.

The reviewer compared all moved class bodies and traced the acyclic import graph, public type re-exports, private compiler-helper import, consistent class identity across alias consumers, SQLAlchemy compiler dispatch/type caches/traversal, hatch package inclusion, existing autoclass/docstring import paths, and exact-once test/helper discovery. No defect found. Existing public type names remain importable from types.py, while ARRAY implementation and its supporting classes now live together in array.py.

Recorded limits: class __module__ values now name the new modules; previous-module pickles resolve through re-exports, but new pickles need the new modules. Explicit __all__ excludes incidental imported implementation names from wildcard imports. Private helpers moved within this unreleased series. These are not promises of compatibility for undocumented private names or every cross-version serialized object. This remains static review; author runtime/packaging checks are recorded separately and final-head CI is pending.

pyathena/sqlalchemy/types.py
laughingman7743: Self-review round one: CLEAN. Series base 47024e2084ce5c9a338b190d1934151e22c19817, head e6bc6fa6ca8692ffdca1220e95adfda525641a64; bounded comparison ad5e526cd7fbcc504b4a2667a449997a1deb2958..e6bc6fa6ca8692ffdca1220e95adfda525641a64 and affected pytest contracts.

Only test_array.py changes. The four groups are TestAthenaArray (public type construction/reflection/bind-expression contract), TestArrayTypeInspector (element type and unknown-type behavior through public compilation hooks), TestArrayValueProcessor (bind/literal/result conversion), and TestArrayJSONProjection (outer projection and native ordering). Five helper type classes remain module-level and precede decorators that instantiate them. No class has setup, mutable class state or a custom constructor.

A normalized AST comparison confirms all 47 test bodies, arguments and decorators are identical except for adding self to former functions; helper class ASTs are unchanged. Required format/lint and the same 164 type/compiler cases pass.

docs/sqlalchemy.md
laughingman7743: Self-review round two: CLEAN. Series base 47024e2084ce5c9a338b190d1934151e22c19817, head e6bc6fa6ca8692ffdca1220e95adfda525641a64; bounded comparison ad5e526cd7fbcc504b4a2667a449997a1deb2958..e6bc6fa6ca8692ffdca1220e95adfda525641a64 and affected pytest contracts.

The explicit request to group this file by implementation class takes precedence over the usual standalone-unit-test convention. Tests continue exercising observable public type and compiler behavior; no private-state assertions, new mocking layer or production changes were added. Parametrization still supplies the original named arguments, and pytest binds self normally. Test identifiers gain class names, but existing function names and -k selection remain available. No repository command or documentation pins the moved standalone node IDs. Earlier module/import/docs/wheel validation remains applicable because the production tree is identical. Independent grouping review and current-head CI remain pending.

tests/pyathena/sqlalchemy/test_array.py
```diff
@@ -0,0 +1,576 @@
...
+            "array(row(integer, varchar))",
+        ],
+    )
+    def test_array_reflection_warns_for_unrecognized_nested_type(self, signature):
+        with pytest.warns(sa_exc.SAWarning, match="Did not recognize type"):
+            type_ = AthenaDialect()._get_column_type(signature)
+        assert isinstance(type_, types.NullType)
+
+
+class TestArrayTypeInspector:
```
laughingman7743, line 211: Relayed independent test-grouping review: CLEAN. Claude claude-fable-5-1, verified first-party Max profile max, effort high; session a55b4852-cae1-4927-8b9b-0bef13ae442d. Series base 47024e2084ce5c9a338b190d1934151e22c19817; previous head ad5e526cd7fbcc504b4a2667a449997a1deb2958; current head e6bc6fa6ca8692ffdca1220e95adfda525641a64. Literal test-only repair/series comparison plus directly affected implementation classes, pytest configuration and fixtures. Read/Glob/Grep only; no execution, network, edits or PR discussion. Actual result collected; snapshot and PR worktree unchanged.

The reviewer verified all 37 former functions plus 10 existing methods are collected through four Test-prefixed classes without constructors or duplicate method names; every parameter table, stacked decorator order, assertion and test value remains unchanged. Five support classes are defined before decorators use them, have no Test prefix, and retain their previous behavior. No test ID references or dependency markers require updates. Each test constructs its own dialect and introduces no shared class state.

Grouping follows the four implementation responsibilities through public behavior. Some pre-existing cases span compiler/reflection/formatter boundaries, so their intact placement is a judgment call rather than isolated testing of one private method; the reviewer found no action needed. The author separately verified normalized AST equivalence and ran the same 164 cases successfully. Static review does not establish xdist execution; final-head CI is running.

laughingman7743: Final validation for e6bc6fa6ca8692ffdca1220e95adfda525641a64: All 18 current-head CI checks pass: ordinary, synchronous SQLAlchemy and asynchronous SQLAlchemy tests on Python 3.10–3.14, both docs checks, and Benchmark tooling. Test run: https://github.com/pyathena-dev/PyAthena/actions/runs/35519323574.

The published head matches the reviewed head; the worktree is clean and the merge base remains 47024e2084ce5c9a338b190d1934151e22c19817. GitHub reports MERGEABLE. Both self-review perspectives and the collected independent review are complete. All readiness checks have passed.

docs/sqlalchemy.md
```diff
@@ -1025,90 +1078,72 @@ This creates a table definition equivalent to:
...
-from sqlalchemy import create_engine, select
-
-# Query ARRAY data using ARRAY constructor
-result = connection.execute(
-    select().from_statement(
-        text("SELECT ARRAY[1, 2, 3, 4, 5] as item_ids")
-    )
-).fetchone()
+Use `select()` with `ARRAY` or `AthenaArray` columns whose element types are known, either declared explicitly or reflected from Athena, to receive typed Python collections.
+For example, the `events` table above returns `numbers` as a list of integers and `labels` as nested lists of strings.
```
laughingman7743, line 1090: Documentation repair `e6bc6fa6ca8692ffdca1220e95adfda525641a64..dea235f30903d2a7fc50b22e9a35e205bda4dcdc`; frozen merge base `47024e2084ce5c9a338b190d1934151e22c19817`. Initial full reviews remain recorded. Round one follow-up: CLEAN. Checked all changed ARRAY prose/examples against DefaultTypeConverter, native ARRAY parsing, result-set dispatch, and the typed ARRAY projection/result processor. The four table examples and JSON decoding were verified locally; the typed SELECT compiles to the JSON transport projection. No runtime source or tests changed. `just format`, `just lint`, `just docs lint`, and a current-source Sphinx HTML build pass. No AWS queries were run for this documentation repair.

docs/sqlalchemy.md
```diff
@@ -1025,90 +1078,72 @@ This creates a table definition equivalent to:
...
+The unquoted nested string array is not valid JSON, and the native parser does not support nested arrays.
+The last example contains an equals sign that the native parser rejects.
+For these unsupported representations, the converter returns the original string rather than dropping unparsed elements.
+A string result therefore does not necessarily mean the stored ARRAY is invalid.
+Calling `json.loads()` on that native text will not resolve these cases because it is not JSON.
 
-```python
-# Input: '[1, 2, 3]'
-# Output: [1, 2, 3]
+Use typed SQLAlchemy SELECT expressions when element types and nested string values must be preserved.
```
laughingman7743: Documentation repair `e6bc6fa6ca8692ffdca1220e95adfda525641a64..dea235f30903d2a7fc50b22e9a35e205bda4dcdc`; frozen merge base `47024e2084ce5c9a338b190d1934151e22c19817`. Initial full reviews remain recorded. Round two follow-up: CLEAN. The direct-query examples explicitly require the standard REST cursor/default converter with no custom conversion or result type hints. Existing simple ARRAY conversion is distinguished from the new typed SQLAlchemy path; unsupported native formats are distinguished from invalid stored values. Removed the ARRAY raw-string migration example and redundant JSON parsing. Declared-type conversion and JSON-derived values are separated. Reviewed the rendered section and NULL/empty-array example. Other cursor implementations are explicitly outside these examples. Current-head CI is pending.

laughingman7743: Relayed independent documentation review: **FINDINGS**, with no factual errors found in the new text. Claude `claude-fable-5-1`, first-party `max`, effort `high`; session `033646dc-9673-41bb-b3de-c668b85a30f1`. Static read-only review of `e6bc6fa6ca8692ffdca1220e95adfda525641a64..dea235f30903d2a7fc50b22e9a35e205bda4dcdc`, merge base `47024e2084ce5c9a338b190d1934151e22c19817`, using four small public diff/context files. Actual result collected; package and PR worktree remained unchanged.

The reviewer identified one clarity gap: native parsing may return a list with incorrect element boundaries, so a list result is not proof of faithful conversion. Three minor suggestions concern equals signs inside ROW elements, attributing the JSON projection to PyAthena, and showing where `cursor` comes from. The follow-up adds a flat string-array example and a comma/NULL ambiguity example, narrows the equals-sign explanation, attributes projection to PyAthena, and links cursor setup. Local inspection confirmed `_convert_value` preserves `001` as a string in the native fallback, so the reviewer's tentative numeric inference is not adopted.

Separately, restoring the existing `Querying ARRAY data` heading resolves an adjacent insertion conflict in the dependent PR while retaining the clarified body. These repairs will receive both self-review perspectives and a bounded independent follow-up. No AWS execution was performed by the reviewer.

docs/sqlalchemy.md
laughingman7743: Final documentation repair `dea235f30903d2a7fc50b22e9a35e205bda4dcdc..50f07479bf2eb4e0a8fb9cbad8274c263e1be671`; merge base `47024e2084ce5c9a338b190d1934151e22c19817`. Round one follow-up: CLEAN. Verified the flat string and comma/NULL ambiguity examples against DefaultTypeConverter; `_convert_value` preserves numeric-looking native strings and the docs do not claim otherwise. The ROW equals-sign exception, PyAthena projection attribution and cursor setup link match source. Restoring the original section heading merges cleanly with the dependent expression section. Only docs/sqlalchemy.md differs from the previously validated implementation. Format, lint, Markdown lint and current-source HTML build pass.

docs/sqlalchemy.md
laughingman7743: Final documentation repair `dea235f30903d2a7fc50b22e9a35e205bda4dcdc..50f07479bf2eb4e0a8fb9cbad8274c263e1be671`; merge base `47024e2084ce5c9a338b190d1934151e22c19817`. Round two follow-up: CLEAN. A successful list conversion is now explicitly distinguished from faithful element recovery; the examples no longer suggest testing str versus list is sufficient. Guidance covers string elements and nested arrays before native parsing can misinterpret them. The usage link resolves in rendered HTML, with no warning in the final incremental build. The ordinary cursor behavior remains scoped to default REST conversion. Converter/example checks were local and did not execute AWS queries; latest CI is still pending.

laughingman7743: Relayed final documentation follow-up: **CLEAN**. Claude `claude-fable-5-1`, first-party `max` profile, effort `high`; session `7a580d42-8de2-4dd6-ab72-b0671c77b0c3`. Static read-only review of `dea235f30903d2a7fc50b22e9a35e205bda4dcdc..50f07479bf2eb4e0a8fb9cbad8274c263e1be671` with merge base `47024e2084ce5c9a338b190d1934151e22c19817`. The actual result was collected; package hashes and the clean PR worktree were unchanged.

Confirmed the restored heading, cursor scope, flat-string conversion, ROW-specific equals-sign explanation, comma/NULL ambiguity, PyAthena projection attribution and guidance to use typed SELECT or JSON for strings/nested arrays. Optional suggestions to spell out the flat-string parser path and add numeric/boolean-looking string examples are deferred: the preceding JSON/native description and the general guidance already cover those cases.

Package limits checked by the author: the usage page exists and the link renders correctly; the old pre-repair heading anchor has no references in project docs. Final Markdown lint and current-source HTML build pass, and the added converter examples match local results. The reviewer did not execute tests or AWS queries. Latest CI is still pending.

laughingman7743: Final documentation validation for `50f07479bf2eb4e0a8fb9cbad8274c263e1be671`: **all 18 current-head CI checks succeeded**, including ordinary, synchronous SQLAlchemy and asynchronous SQLAlchemy tests on Python 3.10–3.14, docs lint/build and benchmark tooling. [Test run](https://github.com/pyathena-dev/PyAthena/actions/runs/35553777307). Both self-review perspectives and the collected independent CLEAN follow-up cover this revision. Local/published/reviewed heads match, the worktree is clean, and GitHub reports MERGEABLE. The dependent expression PR also merges cleanly after preserving the original heading. Restoring Ready; no merge performed.

PR #775, opened 2026-09-19
Support SQLAlchemy ARRAY expressions in SELECT and WHERE

laughingman7743:
## WHAT

Compile ARRAY indexing, inclusive slices, concatenation, and ANY/ALL into Athena expressions for SELECT and WHERE.
Preserve dimensions, zero-based index translation, bound values, SQL NULL semantics, comparison direction, and SQLAlchemy FROM linting.
Avoid capturing column names in lambda arguments, render read-index expressions once, and reject unsupported slice steps even when a compiled statement is reused.

Keep the comparator and slice-step type in `array.py`, resolving decorated and variant ARRAY implementations through the shared type inspector.
Group comparator tests by responsibility in `test_array.py` and SQL generation tests in the existing compiler test class.

Read indices outside the valid SQL range return NULL; slice stops are inclusive, reversed slices are empty, and only None/1 steps are accepted.
Slice array and boundary expressions must be deterministic because generated SQL can evaluate them more than once.
Explicitly group quantified comparisons before negating them; SQLAlchemy can otherwise invert the element comparison.
Use AthenaArray for open-ended zero-indexed slices; SQLAlchemy's generic ARRAY comparator requires explicit boundaries in that mode.
ANY/ALL over subqueries retain their normal compilation.

## WHY

Native ARRAY values need SELECT and WHERE operations.
Direct Athena subscripts can fail for out-of-range indices, and array ANY/ALL comparisons need lambda expressions.

Refs #755. Builds on merged #774. Partial UPDATE support follows in #776.

## TEST

Rebased revision: `fbcf8b88d845eb7577d5c481fad0fb265e832ebb`, based on `master` at `08e97577224a05e32653b6a3e34ff76f6e571e89`.
Local environment: Python 3.13.1, SQLAlchemy 2.0.46.

- `just format`, `just lint`, and `just docs lint` pass.
- The focused command below passes 215 tests without AWS queries. An initial invocation omitted `.env` and failed during collection; loading the linked `.env` resolved it without a code change.
- `uv run sphinx-build -b html docs /private/tmp/pyathena-array-775-postmerge/html` builds the current source documentation; existing Sphinx warnings remain.
- Source and test trees are identical to reviewed revision `d3b296cdfa63dff4b5cd06b4949324a810b19684`, which passed 8 targeted real-Athena expression cases in each sync/async mode, SQLAlchemy 2.0.0 compatibility smoke, and all 18 CI checks. Those are prior-revision results; current-head CI passed all 18 checks. Both self-review perspectives and the user-selected `claude-opus-5` (`max`, effort `high`) rebase follow-up are complete with a CLEAN result, recorded inline.

```bash
uv run --env-file .env pytest --noconftest \
  tests/pyathena/sqlalchemy/test_types.py \
  tests/pyathena/sqlalchemy/test_array.py \
  tests/pyathena/sqlalchemy/test_map.py \
  tests/pyathena/sqlalchemy/test_struct.py \
  tests/pyathena/sqlalchemy/test_temporal.py \
  tests/pyathena/sqlalchemy/test_compiler.py -q
```


pyathena/sqlalchemy/compiler.py
```diff
@@ -259,6 +263,130 @@ def _array_type_inspector(self):
...
         return f"length{self.function_argspec(fn, **kw)}"
 
+    @staticmethod
+    def _original_froms(elements):
+        for element in elements:
+            while element._is_clone_of is not None:
+                element = element._is_clone_of
+            yield element
+
+    def _array_lambda_name(self):
```
laughingman7743, line 273: Self-review round 1 (implementation behavior): FINDINGS. Full merge-base e6bc6fa6ca8692ffdca1220e95adfda525641a64 to published head 4bebcf886365a4cd0aa3d546483486689550f691. Covered array.py, compiler.py, both type/compiler test files, the SQLAlchemy integration suite, and the ARRAY documentation. Traced comparator dimensions/zero indexes, decorated and variant types, native projection callers, cache/step validation, comparison flipping and FROM linting, and the supported dependency floor. Local evidence: 208 pure cases, 17 sync and 17 async ARRAY cases, plus the SQLAlchemy 2.0.0 smoke. Two compiler edge cases need repair before readiness.

[P2] Avoid capturing identifiers embedded in SQL fragments. literal_column("_pyathena_element_0 + 1") == any_(items) currently compiles to any_match((items), _pyathena_element_0 -> _pyathena_element_0 + 1 = _pyathena_element_0). The real column reference becomes the lambda element, so the predicate compares each element plus one to itself. Checking only exact node names misses identifiers inside literal_column and TextClause fragments. Include those fragments when choosing a fresh lambda name and cover the actual WHERE result.

pyathena/sqlalchemy/compiler.py
```diff
@@ -259,6 +263,130 @@ def _array_type_inspector(self):
...
+            sql = f"slice({array}, {start}, {length})"
+            return self._array_slice_step(sql, bounds.step, array_type, **kw)
+        index_expression = binary.right
+        if (
+            isinstance(index_expression, BindParameter)
+            and self._array_type_inspector.array_type(index_expression.type) is not None
+        ):
+            index_expression = index_expression._with_binary_element_type(types.Integer())
+        index = self.process(index_expression, **kw)
+        return f"element_at({array}, IF({index} > 0, {index}, NULL))"
```
laughingman7743: Self-review round 1 finding at base e6bc6fa6ca8692ffdca1220e95adfda525641a64, head 4bebcf886365a4cd0aa3d546483486689550f691. [P2] Render an index expression once. IF(index > 0, index, NULL) duplicates the caller expression. With a volatile integer index, the positive guard and the selected index can disagree: a second negative value accesses from the end, or zero raises, despite the NULL contract. Use a single-occurrence expression to map nonpositive indices to NULL, retain the NULL-index case, and assert that a volatile index appears once in compiled SQL.

pyathena/sqlalchemy/compiler.py
laughingman7743: Round-one repair follow-up: CLEAN for 4bebcf886365a4cd0aa3d546483486689550f691..267663211698c06068c3c3dc023db266c62ea4ca, with series base e6bc6fa6ca8692ffdca1220e95adfda525641a64. The name check now includes TextClause fragments and identifiers embedded in literal SQL expressions, using a conservative substring check before allocating a lambda name. The original Column case plus raw, quoted and TextClause expressions pass. The real-Athena WHERE regression returns row id 2 for the textual operand instead of comparing the lambda element to itself. All 212 pure cases and all 6 current expression cases pass synchronously and asynchronously with -n 1. The inherited type/value processors and the SQLAlchemy 2.0.0 smoke remain covered. Complete current-head CI is running.

pyathena/sqlalchemy/compiler.py
laughingman7743: Round-one repair follow-up: CLEAN for 4bebcf886365a4cd0aa3d546483486689550f691..267663211698c06068c3c3dc023db266c62ea4ca, with series base e6bc6fa6ca8692ffdca1220e95adfda525641a64. The generated index is now NULLIF(greatest(index, 0), 0), so the caller expression occurs once and nonpositive/NULL indices become NULL. A volatile-index compile regression confirms one random() occurrence; real-Athena coverage includes negative, zero, out-of-range, explicitly NULL, bound, zero-based and decorated indices. All 212 pure cases and 6 sync/6 async expression cases pass. SQLAlchemy 2.0.0 compilation also passes after this repair. Complete current-head CI is running.

docs/sqlalchemy.md
```diff
@@ -1084,6 +1084,36 @@ CREATE TABLE orders (
 )
 ```
 
+#### ARRAY expressions
+
+Use SQLAlchemy expressions to index, slice, concatenate, or compare array values in SELECT and WHERE clauses.
+Indices are one-based by default.
+`AthenaArray(Integer, zero_indexes=True)` translates explicit indices and slice boundaries by adding one.
+Reads return NULL when the resulting SQL index is NULL, below one, or out of range; negative indices do not count from the end.
```
laughingman7743, line 1092: Self-review round 2 (claims, caller compatibility, and operational effects): CLEAN for full merge-base e6bc6fa6ca8692ffdca1220e95adfda525641a64 to published head 267663211698c06068c3c3dc023db266c62ea4ca.

Audited the PR explanation, changed docstrings/comments, the complete six-file diff, and related callers. The documented default and translated indices match comparator construction and NULLIF/greatest rendering; inclusive/clipped/open/reversed slices, retained dimensions and decorators, cached step rejection, all six comparison operators, flipping/negation, and NULL/empty-array outcomes have focused coverage. Literal SQL names are included in lambda allocation, and subquery quantifier operands retain their existing compilation. The documented example uses the preceding orders table and compiles.

The SQLAlchemy 2.0.0 import failure was reproduced and repaired with a TYPE_CHECKING comparator alias; the final compiler repairs also pass the 2.0.0 smoke. Public types.py imports remain intact. Type resolution stays within the dialect's type implementation machinery; this diff changes no AWS metadata calls, credentials, retries, or fixture lifecycle. Integration cases reuse the existing connection/metadata fixtures, and test grouping follows the requested implementation responsibilities.

Evidence is scoped: 212 pure cases and the current six expression cases pass on both sync and async paths; earlier full focused ARRAY runs passed 17 on each path before the final compiler repair. Markdown lint, example compilation and HTML generation have passed; the aggregate current-head CI is still running. The ARRAY function reference was checked against https://trino.io/docs/current/functions/array.html and the stated NULL outcomes were exercised on Athena. No performance or universal SQL-expression coverage claim is made. Ready remains pending the collected independent review and all current checks.

pyathena/sqlalchemy/compiler.py
```diff
@@ -259,6 +263,133 @@ def _array_type_inspector(self):
...
+        )
+        if array_type is not None:
+            variable = self._array_lambda_name()
+            predicate = binary._clone()
+            item_type = _ArrayTypeInspector.item_type(array_type)
+            if aggregate_on_left:
+                predicate.left = Column(variable, item_type)
+            else:
+                predicate.right = Column(variable, item_type)
+                if isinstance(predicate.left, BindParameter) and isinstance(item_type, types.ARRAY):
```
laughingman7743: Independent review (relayed): FINDINGS. Claude claude-fable-5-1, Max profile, effort high; session d7c318ea-c5fe-4c99-8380-b3d8f0634025. Full static review of e6bc6fa6ca8692ffdca1220e95adfda525641a64..267663211698c06068c3c3dc023db266c62ea4ca covered all six changed files, comparator/type resolution, compiler visitors and callers, cache behavior, tests, and documentation. Read/Glob/Grep only; no commands, tests, network, edits, or PR discussion access. Snapshot and PR worktree remained unchanged.

Confirmed by a local compilation probe: String().with_variant(AthenaArray(String), "awsathena") compared through any_(column) == "a" leaves the scalar bind with an ARRAY implementation and fails literal rendering; the corresponding Integer variant also fails. Existing coverage used a mismatched scalar/base type and missed this coercion path. Retype the scalar bind to the resolved array item type when its current dialect implementation is ARRAY. Repair and regression coverage follow. Reviewer found no blocking defects; this concrete failure will nevertheless be repaired before Ready.

pyathena/sqlalchemy/compiler.py
```diff
@@ -259,6 +263,133 @@ def _array_type_inspector(self):
...
+                and bounds.step.value == 1
+            ):
+                raise exc.CompileError("Athena ARRAY slices support only step=None or step=1")
+            start = "1" if isinstance(bounds.start, Null) else self.process(bounds.start, **kw)
+            stop = (
+                f"cardinality({array})"
+                if isinstance(bounds.stop, Null)
+                else self.process(bounds.stop, **kw)
+            )
+            start = f"greatest({start}, 1)"
```
laughingman7743, line 369: Independent review (relayed): low-severity limitation confirmed by compilation. Slice SQL repeats its array/start expressions; items[cast(floor(random()*3), Integer)+1:3] renders random() twice, so volatile boundaries can disagree. Read indexing was already repaired to render its index expression once, but slicing has no equivalent guarantee. This PR will explicitly restrict the documented slice usage to deterministic array/bound expressions; a general expression-hoisting rewrite is deferred because it changes SQL scope/correlation behavior beyond this repair.

docs/sqlalchemy.md
```diff
@@ -1084,6 +1084,36 @@ CREATE TABLE orders (
...
+A reversed slice returns an empty array, and slicing a NULL array returns NULL.
+Only `step=None` and `step=1` are supported.
+Use `AthenaArray` for open-ended slices with `zero_indexes=True`; SQLAlchemy's generic `ARRAY` comparator requires explicit bounds in that mode.
+Indexing a multidimensional array retains the remaining dimensions, while slicing retains its array type.
+
+`any_(array)` and `all_(array)`, and the legacy `array.any(value)` and `array.all(value)` methods, compile to Athena's `any_match` and `all_match` functions.
+Comparisons use SQL three-valued logic: NULL elements can produce NULL when no decisive true or false result exists.
+For an empty array, ANY is false and ALL is true.
+A NULL array produces NULL for both.
+SQLAlchemy comparison flipping and negation are preserved.
```
laughingman7743: Independent review (relayed): the unconditional negation claim is inaccurate for boolean quantifiers. Confirmed by compilation: ~(any_(flags) == True) becomes any_match(flags, e -> e != true), which is not NOT any_match(flags, e -> e = true) for mixed [True, False]. SQLAlchemy performs this inversion before dialect compilation, so it cannot be distinguished from explicit != here. Document the limitation and verify explicit grouping before negation as the supported expression. Runtime results remain unverified for the new regression cases at this point.

pyathena/sqlalchemy/compiler.py
laughingman7743: Self-review round one: CLEAN within the repaired scope. Repair follow-up at da595acefa843edd2b596fd61ece87e8d85239ad, base e6bc6fa6ca8692ffdca1220e95adfda525641a64; compared the old/new patch series and the bounded 267663211698c06068c3c3dc023db266c62ea4ca..da595acefa843edd2b596fd61ece87e8d85239ad repair. Checked all five repair files and direct type-coercion/cache/caller paths. Scalar binds with ARRAY dialect implementations now use the resolved element type; multidimensional ARRAY binds retain their existing treatment. Regression coverage includes String and Integer variants whose base type matches the scalar, literal rendering, repeated bound execution, and grouped versus ungrouped boolean negation. just format/lint and 214 pure cases pass; 8 synchronous and 8 asynchronous ARRAY expression cases pass on real Athena with -n 1. SQLAlchemy 2.0.0 compilation also confirms the repaired variant binds and grouped boolean NOT. No additional actionable implementation findings. Slice volatility remains an explicitly documented restriction, not a claimed fix.

docs/sqlalchemy.md
laughingman7743: Self-review round two: CLEAN within the repaired claims/caller scope. Repair follow-up at da595acefa843edd2b596fd61ece87e8d85239ad, base e6bc6fa6ca8692ffdca1220e95adfda525641a64; compared the old/new patch series and the bounded 267663211698c06068c3c3dc023db266c62ea4ca..da595acefa843edd2b596fd61ece87e8d85239ad repair. Audited the revised negation and slicing claims separately from implementation review: volatile slice inputs are explicitly unsupported; upstream boolean inversion is preserved and the documented self_group() expression generates NOT around the whole match, with [True, False] returning False on both runtime paths. PR validation now distinguishes current 214 pure / 8 sync / 8 async repair results from earlier full focused runs. The compiler change affects only ARRAY-typed scalar-bind coercion; no metadata, retries, account selection, or fixture lifecycle changed. Current aggregate CI and independent follow-up are still pending.

pyathena/sqlalchemy/compiler.py
laughingman7743: Confirmed and repaired in da595acefa843edd2b596fd61ece87e8d85239ad. The scalar bind is retyped when its dialect implementation resolves to ARRAY, preserving the original bind key for statement caching. Added matching-base String and Integer variant cases, including two executions on synchronous and asynchronous Athena; all 8 expression cases pass on each path. 214 pure cases and SQLAlchemy 2.0.0 compilation pass. Both author review perspectives were reapplied; independent follow-up is pending.

pyathena/sqlalchemy/compiler.py
laughingman7743: Disposition at da595acefa843edd2b596fd61ece87e8d85239ad: retained as a documented limitation. The ARRAY expression section now requires deterministic array and boundary expressions for slices and explicitly states they can be evaluated more than once. A general hoisting rewrite is deferred because it adds scope/correlation semantics to this repair. No claim of single evaluation or runtime validation for volatile slicing is made.

docs/sqlalchemy.md
laughingman7743: Corrected in da595acefa843edd2b596fd61ece87e8d85239ad. Documentation explains upstream boolean comparison inversion and shows explicit self_group() before negation. Compilation tests assert both SQL forms; real synchronous and asynchronous tests show ungrouped negation returns True and grouped NOT returns False for [True, False]. The grouped example also compiles with SQLAlchemy 2.0.0.

pyathena/sqlalchemy/compiler.py
laughingman7743: Independent follow-up (relayed): FINDINGS, no blocking defect reported. Claude claude-fable-5-1 / Max / high, session 1c6b9d69-57cc-4aa6-bfb1-c185bec53ae0, reviewed only 267663211698c06068c3c3dc023db266c62ea4ca..da595acefa843edd2b596fd61ece87e8d85239ad plus directly affected coercion/cache/docs/test paths. Static Read/Glob/Grep only; snapshot and PR worktree unchanged. The reviewer verified the target variant fix and grouped boolean semantics, but identified (1) older SQLAlchemy 2.0 also inverts non-boolean quantified comparisons, so the caveat must be general; (2) the broad bind condition can erase an intentionally declared ARRAY type when array_agg understates nesting; and (3) this suite uses sa_testing.combinations rather than pytest.mark.parametrize. The explicit-bind regression is confirmed by a local compilation probe. All three are being addressed; current validation and a further bounded independent follow-up are pending. The initial volatile-slicing limitation remains documented and deferred.

pyathena/sqlalchemy/compiler.py
laughingman7743: Self-review round one repair: CLEAN. At d3b296cdfa63dff4b5cd06b4949324a810b19684 (base e6bc6fa6ca8692ffdca1220e95adfda525641a64), checked the literal old/new series comparison and da595acefa843edd2b596fd61ece87e8d85239ad..d3b296cdfa63dff4b5cd06b4949324a810b19684 repair plus directly affected contracts. The added coercion branch now requires the same type object as the collection operand and non-ARRAY affinity, which is the implicit scalar-affinity variant case. Explicit ARRAY binds retain their declared processors even when array_agg understates the nesting or the collection element type is unknown. Existing multidimensional behavior is unchanged. Regression tests exercise explicit binds, matching-base variants, and generated SQL; integration parameterization now follows sa_testing.combinations. just format/lint, 215 pure cases, and 8 sync/8 async Athena expression cases pass. SQLAlchemy 2.0.0 compilation confirms explicit ARRAY preservation and grouped NOT.

docs/sqlalchemy.md
laughingman7743: Self-review round two repair: CLEAN. At d3b296cdfa63dff4b5cd06b4949324a810b19684 (base e6bc6fa6ca8692ffdca1220e95adfda525641a64), checked the literal old/new series comparison and da595acefa843edd2b596fd61ece87e8d85239ad..d3b296cdfa63dff4b5cd06b4949324a810b19684 repair plus directly affected contracts. The negation caveat now covers version and operand differences rather than only boolean operands. A SQLAlchemy 2.0.0 probe confirms numerical ungrouped negation becomes element-wise != while self_group() retains whole-match NOT. The slice determinism restriction is unchanged. Reviewed PR/docs evidence against the actual 215 pure and 8 sync/8 async repair results; aggregate current-head CI and independent follow-up remain pending. No new AWS calls, retries, metadata behavior, public parameters, or fixture lifecycle changes.

pyathena/sqlalchemy/compiler.py
laughingman7743: All three bounded-follow-up items are addressed in d3b296cdfa63dff4b5cd06b4949324a810b19684: (1) the documentation now recommends grouping for quantifier negation generally, including older 2.0 numerical comparisons; (2) the retyping branch is limited to an implicit same-object scalar-affinity type, preserving explicit ARRAY binds; and (3) integration cases use this suite's sa_testing.combinations. New compile coverage includes shared type objects and unknown element types; synchronous and asynchronous real-Athena tests confirm the array_agg comparison. Both self-review perspectives were reapplied. Further independent review is pending.

laughingman7743: Independent bounded follow-up (relayed): CLEAN. Claude claude-fable-5-1, verified first-party Max profile, effort high; session d78a3b40-d2e2-450b-ba4f-5a2d694323ee. Reviewed da595acefa843edd2b596fd61ece87e8d85239ad..d3b296cdfa63dff4b5cd06b4949324a810b19684 and the corresponding same-base range-diff (base e6bc6fa6ca8692ffdca1220e95adfda525641a64), with direct callers, cache/type inference, tests, and revised documentation in SQLAlchemy 2.0.46 and 2.0.0. Read/Glob/Grep only; no execution, edits, network, PR discussion, or memory access. Actual final result collected; snapshot and worktree unchanged.

The reviewer confirmed that implicit same-object scalar-affinity variant binds are still corrected, explicitly declared ARRAY binds survive the new guard, both new compile assertions fail on the previous implementation, and combinations matches the integration suite convention. Grouped negation and the version-dependent caveat are correct.

Non-blocking limits retained from the review: a possible cache collision requires first executing an invalid ARRAY-versus-scalar statement with an equal-but-distinct variant; a non-identical explicit-variant bind has static coverage only; untyped Python lists against array_agg with understated nesting remain an upstream typing limitation outside this repair. Volatile slice inputs remain excluded by the documented determinism restriction. No further confirmed actionable defects were found in the repair scope.

This is a static independent review. Separately, author validation passes 215 pure cases and 8 synchronous/8 asynchronous real-Athena expression cases, including the new explicit ARRAY comparison. Complete current-head CI is still pending.

laughingman7743: Final CI and readiness verification at d3b296cdfa63dff4b5cd06b4949324a810b19684: all 18 applicable checks completed successfully. Test run https://github.com/pyathena-dev/PyAthena/actions/runs/35526828067 passes all 15 jobs (ordinary, synchronous SQLAlchemy, and asynchronous SQLAlchemy, each on Python 3.10–3.14). Docs Lint 35526827909 and Benchmark tooling 35526827884 also pass. The synchronous Python 3.12 suite reports 529 passed / 766 skipped; these are suite requirement skips, not skipped CI jobs. Published head equals the independently reviewed local head, worktree is clean, base remains feat/755-native-array, and GitHub reports MERGEABLE. Both self-review rounds and the collected Claude independent follow-up are complete. Marking this PR Ready; no merge is performed.

pyathena/sqlalchemy/compiler.py
laughingman7743: Rebase follow-up: `e6bc6fa6ca8692ffdca1220e95adfda525641a64..d3b296cdfa63dff4b5cd06b4949324a810b19684` compared with `08e97577224a05e32653b6a3e34ff76f6e571e89..fbcf8b88d845eb7577d5c481fad0fb265e832ebb`. Initial full reviews remain recorded above. Round one: CLEAN. Audited all six PR paths and the upstream changes. Git object IDs for the complete pyathena and tests trees, pyproject.toml and uv.lock are identical before/after the rebase. The 12-commit range comparison differs only in surrounding documentation context; ARRAY expression behavior and test coverage are unchanged. The merged ARRAY conversion explanation remains intact alongside the expression section. Format/lint and 215 focused type/compiler tests pass with the linked .env; initial collection without .env failed before test execution and was corrected. No new live AWS run was needed before current CI.

docs/sqlalchemy.md
laughingman7743: Rebase follow-up: `e6bc6fa6ca8692ffdca1220e95adfda525641a64..d3b296cdfa63dff4b5cd06b4949324a810b19684` compared with `08e97577224a05e32653b6a3e34ff76f6e571e89..fbcf8b88d845eb7577d5c481fad0fb265e832ebb`. Initial full reviews remain recorded above. Round two: CLEAN. Checked the merged documentation against read/index/slice/quantifier claims, typed transport and direct-cursor limits; rendered sections coexist with valid anchors. Upstream runtime source, tests and dependencies are unchanged. The new CI admission condition allows this same-repository PR; expected ordinary/dialect/async coverage remains required. New contribution guidance is followed: all PR files already exist and the description now uses WHAT/WHY/TEST with exact revision, environment, commands and prior/current evidence boundaries. Current-source Sphinx HTML builds (154 existing warnings); latest CI is pending.

pyathena/sqlalchemy/compiler.py
laughingman7743: Independent rebase follow-up is **incomplete** for `fbcf8b88d845eb7577d5c481fad0fb265e832ebb` (base `08e97577224a05e32653b6a3e34ff76f6e571e89`). The requested `claude-fable-5-1`, first-party `max`, effort `high` invocation returned a Fable usage-limit error before review; session `2e5f5ce6-35a2-4ddb-b9c0-87cc69706c33`. This is not a CLEAN review. The selected account/profile was verified, and the bounded public package and clean worktree remained unchanged. Both author review perspectives and local validation are complete; current CI continues. The PR remains Draft pending the requested reviewer becoming available or an explicit reviewer change.

laughingman7743: Relayed independent rebase review: **CLEAN**. At the user's direction, switched from the unavailable Fable model to `claude-opus-5`, verified first-party `max`, effort `high`. Session `248faa4a-b3e8-4239-a512-6992c49e14e6`. Actual static review result collected for `e6bc6fa6ca8692ffdca1220e95adfda525641a64..d3b296cdfa63dff4b5cd06b4949324a810b19684` versus `08e97577224a05e32653b6a3e34ff76f6e571e89..fbcf8b88d845eb7577d5c481fad0fb265e832ebb`; no commands, edits, tests, network, PR discussion or memory access. Four focused public files were supplied; package hashes and the clean worktree were unchanged.

Covered unchanged expression contracts, the documentation-context range difference, merged ARRAY guidance and updated CI/test admission rules. No findings. Author verification confirms the range difference is surrounding context only; source/tests and dependency object IDs match exactly.

Package limits checked separately by the author: the head and base repositories are both `pyathena-dev/PyAthena`, so the fork guard admits these jobs. `AthenaAioDialect` inherits `AthenaDialect.statement_compiler`, so current validation includes the async PyAthena directory through the ordinary test job plus the separate async compliance jobs. The usage link resolves in the locally built HTML. Current-head CI remains pending; the reviewer did not execute validation. This completes the previously blocked independent-review step.

laughingman7743: Final validation for rebased `fbcf8b88d845eb7577d5c481fad0fb265e832ebb`: all **18 current-head checks succeeded**, including ordinary, synchronous SQLAlchemy, and asynchronous SQLAlchemy jobs on Python 3.10–3.14, documentation lint/build and benchmark tooling. [Test run](https://github.com/pyathena-dev/PyAthena/actions/runs/35557128156). Two self-review perspectives and the collected independent `claude-opus-5` (`max`, high effort) CLEAN result cover this revision. GitHub reports MERGEABLE; published/local/reviewed heads match and the worktree is clean. Restoring Ready without merging.

Issue #780, opened 2026-09-19
Investigate persistent metadata throttling during SQLAlchemy reflection


laughingman7743:
## Problem

SQLAlchemy reflection can still fail with Athena metadata throttling after the request-reuse and concurrency changes in #777.
Restoring reflection coverage in #771 exposed the problem during the compliance suite; the root cause and practical impact outside CI are not yet established.
Track the remaining investigation separately from the weekly database sweep and development-tooling work in #779.
This blocks completion of the prerequisite metadata fix for the ARRAY work in #755.

## Current implementation and evidence

Continue from #777, currently Draft at `ea4ce1fc04c922592ef9f6b0fad46702c894598c`.
Do not discard the existing implementation or start a replacement PR without a concrete reason.

- Positive table metadata is shared within one SQLAlchemy Inspector across listings and column/comment/table-option reflection. Catalog and schema identity, cache invalidation, and preservation of earlier results have regression coverage.
- Fresh CREATE/DROP existence checks remain uncached. A profiled synchronous baseline attributed 378 of 643 GetTableMetadata SDK calls to these checks.
- Wrapped throttling, access denial, and unknown metadata failures must propagate rather than establish or cache table absence.
- CI SDK retries are bounded to standard mode with three attempts. SQLAlchemy matrices run at most two Python versions concurrently, with eight workers per job; this does not limit traffic from other PRs.
- [CI run 35448063098](https://github.com/pyathena-dev/PyAthena/actions/runs/35448063098) completed on that head: all five regular PyAthena jobs passed; SQLAlchemy 3.10, 3.11, and 3.12 failed, while 3.13 and 3.14 passed. The dependent async suite was skipped.
- Catalog cleanup improved database enumeration, but metadata throttling recurred afterward. The database backlog is not established as its cause.
- Earlier full local synchronous and asynchronous baselines passed, but used a different AWS account from CI. They do not establish that the current head passes in the CI account. Use the current local `.env` configuration and verify the target before interpreting new results.
- Independent review findings about error-propagation regression coverage and documentation were repaired in the current head. The final independent follow-up remains incomplete.

## Questions to resolve

1. Which API calls, callers, retry layers, and concurrency patterns cause the remaining failures in the intended test account? Separate Inspector reflection from fresh existence checks and distinguish client calls from actual SDK attempts.
[...]

PR #777, opened 2026-09-19
Fix metadata reflection throttling and reuse listed metadata

laughingman7743:
## WHAT

Share positive table metadata from listings and individual lookups across column, comment, and table-option reflection within one SQLAlchemy Inspector.
Cache identity uses the effective cursor catalog and resolved schema, including cursor catalog overrides and default-schema aliases.
Table-name case variants share results only in `AwsDataCatalog`.
A later listing preserves metadata already fetched for a table; `Inspector.clear_cache()` invalidates both forms.
Listing omissions do not establish absence, and direct CREATE/DROP existence checks remain fresh.

Recognize Glue error codes wrapped in Athena `MetadataException` responses for configured retries and missing-table handling.
Table-metadata throttling, permission failures, and unknown errors propagate instead of becoming false missing-table results.
Capture retry exception names once so tuple, list, iterator, and single-string policies work across repeated attempts and API calls.

Make PyAthena's API retries outlast metadata throttling episodes:

- `retry_api_call` adds uniform jitter of up to one `multiplier` to each exponential wait, so concurrent clients stop retrying in lockstep.
- The default `RetryConfig` makes eight attempts instead of five; the exponential waits sum to 127 seconds plus jitter instead of 15 seconds.
- `is_retryable_error()` is exposed and shared by the retry loop and the dialect.

Column reflection and `has_table()` make one table-metadata attempt; when Athena reports `ThrottlingException` or `TooManyRequestsException`, they read `information_schema.columns` (column name, type, comment, partition marker) at once instead of waiting out the retry policy.
The query runs without result reuse and with plain equality filters, and its columns are cached under their own Inspector key so they are never mistaken for full table metadata.
Table comments and table options still come from the metadata API with the configured retries, so they propagate the throttling error rather than returning partial metadata.
Permission failures and unrecognized errors still propagate from `has_table()`.
Other codes in `RetryConfig.exceptions` are still retried on that path with the connection's policy; only the throttling codes and `MetadataException` itself are removed. The async connection adapter forwards cursor keyword arguments and `retry_config` so both dialects share the path.

CI keeps SDK retries at standard mode with three attempts and runs all five Python versions concurrently again, so the CI suites exercise PyAthena's own retry policy and fallback under real contention. The reusable workflow input for the matrix parallelism stays available.

## WHY

Reflection can reuse metadata already returned by Athena, reducing duplicate requests without adding a stale connection-wide cache.
The retry/error changes preserve the distinction between a missing table and a failed metadata request.

[...]

PR #782, opened 2026-09-20
Reflect an empty information_schema comment as no comment

### LINKED ISSUE #781: Verify metadata error handling and the information_schema fallback on federated catalogs and DataFrame cursors
## Problem

PR #777 (#780) changed how SQLAlchemy reflection decides between "table does not exist" and "the metadata request failed", and added an `information_schema.columns` fallback for `has_table()` and column reflection when Athena throttles `GetTableMetadata`.
Both behaviors were verified against the Glue Data Catalog (`AwsDataCatalog`) and, for listing semantics, an S3 Tables catalog.
They are not verified against Lambda-based or Hive-metastore federated catalogs, or through the DataFrame/UNLOAD cursor classes.

## What is unverified

1. **Missing-table detection in federated catalogs.** `pyathena/sqlalchemy/base.py` maps a `MetadataException` to `NoSuchTableError` only when its message carries the `(Service: AmazonDataCatalog; ...; Error Code: EntityNotFoundException; ...)` envelope (`pyathena/util.py`, `_get_error_code(..., unwrap_metadata=True)`). Before #777 any `MetadataException` meant "missing". If a federated catalog reports a missing table with a different message, `has_table()` now raises `OperationalError` instead of returning `False`, and `create_all(checkfirst=True)` fails.
2. **`information_schema.columns` in federated catalogs.** The fallback assumes the query is served for the cursor's catalog, that names are reported in lowercase, and that `extra_info = 'partition key'` marks partition columns.
3. **`ListTableMetadata` `Expression` for names over 128 characters.** The lookup lists with a lowercase regex and matches case-insensitively; the filter semantics of non-Glue catalogs are unknown.
4. **DataFrame/UNLOAD-backed cursors.** The fallback runs through the connection's cursor class. NaN-comment normalization and client-side ordering by `ordinal_position` are pinned by a stub-cursor unit test (`tests/pyathena/sqlalchemy/test_base.py::TestAthenaDialect::test_columns_from_information_schema`), but no live test runs the fallback through `awsathena+pandas`, `awsathena+arrow`, or `awsathena+polars`, with or without `unload=true`.

## Proposed verification

- Item 4 needs no new infrastructure: add one live case equivalent to `tests/sqlalchemy/test_suite.py::HasTableTest::test_throttled_metadata_requests_use_information_schema` to `tests/pyathena/sqlalchemy/test_base.py` for the DataFrame dialects (force a `ThrottlingException` from `get_table_metadata`, assert the reflected columns, comments and partition flags).
- Items 1-3 need a federated catalog in the test account (an Athena Lambda connector or a Hive metastore connector). Record the raw `MetadataException` message for a missing table, whether `information_schema.columns` returns rows with the expected `extra_info`, and how `ListTableMetadata` treats `Expression`. If the missing-table envelope differs, decide whether to recognize the additional format explicitly rather than restoring the blanket `MetadataException` mapping, which turned throttling into false missing-table results.

## Context

- #777 merged as `7964872`; findings and evidence are recorded inline there and in #780.
- `docs/sqlalchemy.md` documents the current behavior; update it once federated catalogs are verified.

laughingman7743:
## WHAT

Verify the `information_schema` fallback added in #777 through the DataFrame-backed SQLAlchemy dialects against live Athena, and repair what that verification found.

`tests/pyathena/sqlalchemy/test_base.py::TestSQLAlchemyAthena::test_throttled_columns_across_cursor_types` forces a `ThrottlingException` from `GetTableMetadata` on the connection under test, then reflects a table with a commented column, an uncommented column and a partition column. It runs for `awsathena+pandas`, `awsathena+arrow` and `awsathena+polars`, each with and without `unload=true` — six engines. It asserts the reflected column order (the fallback query has no `ORDER BY`, so the order comes from the client-side `ordinal_position` sort), the comments, the partition flags, the reflected types, and — through a `provide-client-params.athena.StartQueryExecution` hook — that reflection issued the fallback and nothing else, in the execution mode the case asked for.

That test found a real defect. `_columns_from_information_schema` accepted any `str` as a comment, and a CSV-backed arrow cursor reports a NULL comment as `""`, so a column with no comment reflected as `""` where every other path reports `None`. Measured per cursor on the same `information_schema.columns` row:

| cursor | `comment` for a NULL |
|---|---|
| rest | `None` |
| pandas (CSV) | `nan` |
| pandas + unload | `None` |
| arrow (CSV) | `''` |
| arrow + unload | `None` |
| polars (CSV / unload) | `None` |

The repair normalizes in `_column()`, the single construction site both reflection paths funnel through, so a Glue comment of `""` now reflects as `None` on the metadata API path too instead of disagreeing with the fallback. `_columns_from_information_schema` keeps its `isinstance(comment, str)` guard, which is what rejects pandas' `nan` — `_column()` cannot, since `nan` is truthy. `get_table_comment()` gets the same rule: it was the one reflection method still able to report `{"text": ""}` for a table whose comment the DDL compiler would refuse to write back (`compiler.py` skips a falsy comment).

## WHY

#781 records that the #777 fallback was verified only against `AwsDataCatalog` through the `rest` cursor. The implementation comments claimed that UNLOAD-backed cursors need the client-side sort and that cursors reading NULL as NaN need comment normalization, but only a stub-cursor unit test pinned those claims. The live run confirms the sort and the partition marking for every DataFrame cursor, and it disproves the comment normalization as written: NaN was not the only missing-comment shape.

Reflection is expected to agree across cursor classes. A column with no comment reflected as `""` through arrow and as `None` everywhere else is a difference users can observe in `Table.c.<col>.comment` whenever metadata is throttled.

Behavior change beyond the fallback: a column or table whose Glue comment is literally `""` previously reflected as `""` and now reflects as `None`. PyAthena's own DDL compiler cannot produce that state, so a `""` comment does not survive a DDL round trip today either; this only affects comments written by Glue, Spark or another tool.

## Validation

- `just format`, `just lint` (ruff, ruff format, mypy) pass.
- `pytest tests/pyathena/sqlalchemy/test_base.py::TestAthenaDialect` — 6 passed, no AWS. Two are new: the empty-`information_schema`-result branch (`has_table()` reports absence rather than propagating the throttling, and does not cache it) and the empty metadata-API comment.
- Live, local, `-n 1` at the rebased head: `test_throttled_columns_across_cursor_types` (6 engines) and `test_get_table_comment` — 7 passed.
- Both new assertions were checked for vacuity by making them fail. Without the production change the arrow-CSV case fails with `['identifier', '', ''] != ['identifier', None, None]`. Without the `unload` forwarding in `conftest.py`, `engine1`, `engine3` and `engine5` fail on the execution-mode assertion; the other five cases pass without the production change, which is compatibility coverage rather than a defect.
- The per-cursor table above is measured, not inferred: a throwaway probe ran the same `information_schema.columns` query through each cursor class.
- Two self-review rounds, an independent Codex review and two independent follow-ups on the repairs are recorded inline. The independent review's first pass returned an unsupported `CLEAN` and revised to `FINDINGS` when asked to cite the code; both of its actionable items were repaired and re-reviewed.
- CI passed every job (5 regular, 5 `test-sqla`, 5 `test-sqla-async`) at `e5001d2`, which is this branch before the rebase and before the per-case-table change. CI for the rebased head is pending.

Rebased onto `master` after #784 merged. Two of its changes overlap this one, and both were absorbed rather than duplicated: `information_schema`'s unbounded `varchar` is now mapped back to `string`, so this PR's type assertions follow the metadata API's names instead of `information_schema`'s; and `unload` had already been added to `create_engine`'s option allowlist, so that part of this PR is gone.

Limits: items 1-3 of #781 (missing-table detection, `information_schema` availability, and `ListTableMetadata` `Expression` semantics on Lambda/Hive federated catalogs) are **not** covered here. They need a federated catalog provisioned in the test account and moved to #783.

Closes #781. Refs #777, #780, #783, #784


pyathena/sqlalchemy/base.py
```diff
@@ -395,7 +395,8 @@ def _column(self, name: str | None, type_: str, comment: str | None, partition:
             "nullable": True,
             "default": None,
             "autoincrement": False,
-            "comment": comment,
+            # An empty comment is no comment, whichever path reported it.
+            "comment": comment or None,
```
laughingman7743, line 403: **Self-review round one — implementation behavior.** Perspective: behavior and failure paths, data/resource boundaries, framework contracts, test quality. Base `7964872ca50589418b377b7e04ef27cea49549e1`, head `51171d46846b36e005520416f30d16c5ec404ada` for the initial pass, re-run over `190a9f68` and `96e49c27` for the repairs. Surfaces covered: `_column`, `_columns_from_metadata`, `_columns_from_information_schema`, `_get_columns`/`has_table` absence-vs-failure, `get_table_comment`, the shared async adapter path, and the new/modified tests.

Result: **FINDINGS**, both repaired.

1. `pyathena/sqlalchemy/base.py:430` (as first written) — the empty-comment guard lived only in the fallback, so `_columns_from_metadata` still passed a Glue comment of `""` through verbatim. Scenario: a column whose Glue comment is `""` reflects as `""` on the metadata path and `None` on the throttled path — the same cross-path divergence this change exists to remove, inverted. Repaired in `190a9f68` by moving the rule to `_column()`, the single construction site both paths funnel through. The `isinstance(comment, str)` guard stays in the fallback because `float('nan')` is truthy, so `or None` alone cannot reject pandas' NaN.
2. `tests/pyathena/sqlalchemy/test_base.py` — the empty-`information_schema`-result branch (`if not columns: raise NoSuchTableError`) was covered by a live query on each of six engines, but it does not vary by cursor class. Repaired in `190a9f68`: covered deterministically by `test_empty_information_schema_result_is_a_missing_table` with a stub cursor, and dropped from the live test.

Checked and clean: the fallback result is still cached under its own `pyathena_information_schema_columns` key and never seeded into `pyathena_table_metadata`, so `get_table_comment`/`get_table_options` still reach the API and still propagate throttling; absence is not cached; the async adapter shares `_column` unchanged; `extra_info` needs no normalization because every missing-value shape compares unequal to `'partition key'`.

Limit: `just test sqla` and `sqla-async` were not run locally for this head; the CI jobs of the same name are the evidence for them.

pyathena/sqlalchemy/base.py
```diff
@@ -482,7 +484,8 @@ def get_table_comment(
         self, connection: Connection, table_name: str, schema: str | None = None, **kw
     ):
         metadata = self._get_table(connection, table_name, schema=schema, **kw)
-        return {"text": metadata.comment}
+        # An empty comment is no comment here too; the DDL compiler skips one.
+        return {"text": metadata.comment or None}
```
laughingman7743, line 493: **Self-review round two — claims and operational behavior.** Base `7964872ca50589418b377b7e04ef27cea49549e1`, head `96e49c27661f32b3355622145ef76f1a7d2800e9`. Claims audited: the PR body, the three changed comments in this file, the two test comments, and the commit messages.

Result: **FINDINGS**, all corrected.

1. *"a cursor may read a NULL comment as NaN (pandas), as an empty string (arrow), or as None (polars)"* was asserted before it was measured, and it was incomplete. A probe ran the same `information_schema.columns` query through every cursor class: rest `None`, pandas CSV `nan`, pandas+unload `None`, arrow CSV `''`, arrow+unload `None`, polars CSV and polars+unload `None`. So the shape depends on the transport (CSV vs UNLOAD/Parquet), not on the library alone. The PR body now carries that table and labels it measured; the code comment names the CSV pandas case specifically rather than implying a per-library rule.
2. *"UNLOAD does not preserve ORDER BY"* (inherited from #777) is beside the point here: the fallback query has no `ORDER BY` at all, so nothing is being un-preserved. Corrected in the comment above and in the test.
3. Behavior change beyond the fallback was unstated. `comment or None` in `_column()` and in `get_table_comment()` also changes the metadata-API path: a Glue/Spark/Terraform-written comment of `""` now reflects as `None`. `compiler.py` already skips a falsy comment when emitting DDL, so such a comment does not survive a round trip today either — but it is observable in `Table.c.<col>.comment` and `Inspector.get_table_comment()`, and the PR body now says so.

AWS-operator view: the fallback still issues one `information_schema` query per throttled lookup with result reuse off, unchanged by this PR. The live test's cost is 6 engines x (1 CREATE, amortized by `checkfirst`, + 1 fallback query); `get_columns` after `has_table` is a cache hit, not a second query.

Deferred, recorded not silently dropped: the dialect's internal fallback query runs through the *user's* cursor class, which is why this normalization is needed at all. Forcing an API `Cursor` for dialect-internal queries would remove the need, but it changes #777's shipped shape and affects `get_view_definition` too; that belongs in its own issue, not here.

tests/pyathena/sqlalchemy/test_base.py
```diff
@@ -584,6 +632,85 @@ def test_get_columns(self, engine):
...
+            ({"driver": "pandas"}, False),
+            ({"driver": "pandas", "unload": "true"}, True),
+            ({"driver": "arrow"}, False),
+            ({"driver": "arrow", "unload": "true"}, True),
+            ({"driver": "polars"}, False),
+            ({"driver": "polars", "unload": "true"}, True),
+        ],
+        indirect=["engine"],
+    )
+    def test_throttled_columns_across_cursor_types(self, engine, unload, monkeypatch):
```
laughingman7743, line 649: **Independent review — relayed result.** Reviewer: Codex (GPT-5.x) via the repository's Codex rescue agent, a different model from the author. Read-only: no edits, builds, tests, or GitHub access; static source inspection only. Base `7964872ca50589418b377b7e04ef27cea49549e1`, head `96e49c27661f32b3355622145ef76f1a7d2800e9`. Surfaces the reviewer named: comment normalization across both reflection paths, SQLAlchemy consumers and DDL round trips, the async adapter, cache keys, error/absence semantics, cursor CSV/UNLOAD handling, and the test fixtures and assertions.

The first pass returned `CLEAN` with no supporting detail. I asked it to substantiate each conclusion with `file:line` and a mechanism before I would record the review. It then **revised its verdict to FINDINGS**, with one actionable item.

Held, with citations the reviewer supplied:

1. The two guards are complementary, not redundant. `float('nan')` is truthy, so `nan or None` returns `nan`; removing the `isinstance(comment, str)` guard at `base.py:431` would leak a float into `column['comment']`. Pandas' CSV path produces exactly that (`pandas/result_set.py:222`), pinned by `tests/pyathena/pandas/test_cursor.py:1013-1015`.
2. `get_table_comment()` returning `{'text': None}` instead of `{'text': ''}` is an observable API change (`SA/engine/reflection.py:1325`, and `Table.comment` at `:1961`), but the DDL compiler already skips both values (`compiler.py:794`, `:665`), so recompiling a reflected empty comment produces the same DDL. SQLAlchemy's own comment-reflection suite expects `{'text': None}` for absence (`SA/testing/suite/test_reflection.py:977`). No in-repository dependency on the old value.
3. The monkeypatch reaches the client the fallback uses. Traced `inspect(conn)` -> `Inspector.has_table` -> dialect `_get_columns` -> `_raw_connection()` -> `driver_connection.cursor()` -> `common.py:408`. If it missed, the test would fail anyway: the metadata path maps `string` to `types.String` while information_schema's `varchar` maps to `types.VARCHAR`, and the exact-class assertion separates them.
4. The fixed table name is isolated by the per-process random schema (`tests/__init__.py:41-49`), including under `pytest -n 8`, and session teardown drops the schema. A stale table with a different column set would fail the column-name assertion rather than pass silently, and `checkfirst=True` would not repair it — a latent hazard only if the random-schema convention is bypassed.

**FINDING (actionable), repaired in `b8c3af6`:** all six parametrizations asserted the same normalized result and none asserted the execution mode, so removing the `unload` forwarding from `tests/pyathena/conftest.py` would silently run the three nominal UNLOAD cases as CSV and leave every case green. The test now records the fallback query through a `provide-client-params.athena.StartQueryExecution` hook and asserts `UNLOAD (` against the case's intent.

My first repair attempt read the expected mode back from `engine.url.query`, which is self-consistent: with the forwarding removed, both the URL and the query lose UNLOAD and the assertion still passed. Confirmed by running it. The intent is now a separate parametrize value, and removing the forwarding fails `engine1`, `engine3` and `engine5` as it should.

The reviewer also noted, correctly, that five of the six cases pass without the production change; that is compatibility coverage, not a defect. Only arrow-CSV fails without it.

tests/pyathena/sqlalchemy/test_base.py
```diff
@@ -584,6 +632,85 @@ def test_get_columns(self, engine):
...
+            columns = insp.get_columns(table_name, schema=ENV.schema)
+        finally:
+            raw_connection.client.meta.events.unregister(event, record_query)
+
+        # get_columns() reuses the reflected columns, so the fallback ran once,
+        # and it ran in the mode this case asked for. Without this, an unload
+        # option that stopped reaching the cursor would leave every case green
+        # while three of them silently tested CSV twice.
+        (fallback,) = queries
+        assert "FROM information_schema.columns" in fallback
```
laughingman7743: **Self-review of the repair, both perspectives.** Narrow follow-up on `96e49c27661f32b3355622145ef76f1a7d2800e9..b8c3af62a81c8e894d4bfdb898c7457c01fbaa60`, one commit, one file. The round-one and round-two initial passes above stay full passes; this covers only what the repair changed and the contracts it touches.

Behavior: `indirect=["engine"]` keeps `unload` a plain parameter while `engine` still goes through the fixture — the same split `test_create_table_conn_str` already uses in this file. The hook is registered after `Table(...).create(..., checkfirst=True)` and unregistered in a `finally`, so the CREATE and the `checkfirst` existence check (a metadata API call, not `StartQueryExecution`) stay out of `queries`; the `engine` fixture is function-scoped and builds a fresh connection and boto3 client per case, so nothing leaks between cases either way. `(fallback,) = queries` therefore also pins that `get_columns()` after `has_table()` is a cache hit rather than a second billed query — if that ever changes, this unpacking fails, which is the outcome I want.

Claims: the commit message says removing the `unload` forwarding fails `engine1`, `engine3` and `engine5`. Ran it — those three fail on this assertion and the other three pass, so the assertion is not vacuous and the six cases are not interchangeable. `fallback.strip().startswith("UNLOAD (")` matches what `pyathena/formatter.py::wrap_unload` emits (`textwrap.dedent` leaves a leading newline, hence the `strip()`), and a non-unload cursor sends the bare `SELECT`, which is what the failing run printed.

What this does not establish: the assertion proves the query was submitted as UNLOAD, not that the Parquet round trip produced the values asserted further down. Those come from the reflected columns, which is the coverage that matters here, but the distinction is worth stating rather than implying the test verifies the UNLOAD result format itself.

No assertion that existed before the repair was removed or weakened; the repair only adds.

laughingman7743: **Independent follow-up and a second repair.** Relayed from the same Codex reviewer (read-only, static source inspection, no edits/builds/tests/GitHub access).

Its follow-up on `96e49c27..b8c3af62` confirmed the execution-mode repair resolves the original finding — the intent booleans are independent of what the fixture forwards, `indirect=["engine"]` gives the test the literal bool, the hook sits after the CREATE and its `checkfirst` lookup (a metadata API call, not `StartQueryExecution`), the fixture is function-scoped so nothing leaks, and the `UNLOAD (` prefix matches what `formatter.py::wrap_unload` emits — but it raised one new actionable item.

**FINDING (P2), repaired in `e5001d25`:** `(fallback,) = queries` pinned exactly one *API invocation*, not one fallback execution. `common.py` wraps `start_query_execution` in `retry_api_call`, which re-invokes the client method, and each invocation re-emits `provide-client-params` (botocore's own internal retries sit below that hook and would not duplicate it). So a throttled-then-retried `StartQueryExecution` would leave two entries and fail the unpacking with `ValueError` while reflection had in fact returned the right columns — a spurious failure in exactly the throttling conditions this suite runs under. The assertion now requires at least one recorded query and checks that every one of them is the fallback in the expected mode.

The follow-up on `b8c3af62..e5001d25` returned CONFIRMED-RESOLVED on all four questions I put to it: retried attempts carry identical query text (the UNLOAD location is generated once, before `_execute`, so a retry does not mint a new one); the count assertion's other implication — that `get_columns()` after `has_table()` is a cache hit — is still covered by `tests/sqlalchemy/test_suite.py`, where it was already asserted, so it is relocated rather than lost; `assert queries` plus the per-query checks leave no vacuous pass, including the case where reflection never reaches the fallback; and nothing else was weakened.

Re-validated after the change: 6 passed live, and removing the `unload` forwarding from `conftest.py` still fails `engine1`, `engine3` and `engine5`, so the mode assertion did not go slack.

Remaining limit, stated rather than implied: the substring check is not SQL validation — it does not re-verify the schema/table predicates, which the stub-cursor unit test covers instead.

laughingman7743: **Repair from CI, recorded here because it reverses a decision made in the `/simplify` pass above.**

CI for `b8c3af62` failed four of five `test` jobs, and the failures are informative rather than noise:

- 3.10 and 3.12: `test_create_table_with_varchar_text_column` failed on `assert not isinstance(actual.c.col_varchar.type, types.VARCHAR)`. That test asserts the metadata API's type names; when `GetTableMetadata` is throttled the reflection falls back to `information_schema`, where `STRING` is reported as `varchar`. So the failure is the #777 fallback firing under CI load, not a defect in this PR.
- 3.11 and 3.14: `test_cast_as_varchar` failed with `ThrottlingException ... (reached max retries: 2)` on `GetTableMetadata`.

Both point at the same scarce resource. Athena rate-limits the metadata APIs per account, which is the whole premise of #777, and `Table(...).create(bind=conn, checkfirst=True)` spends one `GetTableMetadata` call per parametrized case. The `/simplify` round adopted `checkfirst` with a fixed table name to avoid six identical `CREATE` statements — correct on billed-query count, wrong on the resource that actually constrains this suite. A DDL query costs no metadata capacity; the existence check costs exactly the capacity the test is contending for.

Reverted in `d2f9b68a`: each case creates its own uuid-suffixed table again, with a comment naming the trade so the next simplification pass does not re-take it. Session teardown drops the schema with `CASCADE`, so the extra tables are not leaked.

Not fixed here, and flagged rather than folded in: `test_create_table_with_varchar_text_column` and the other reflection tests that assert metadata-API type names are not throttling-tolerant. That is a pre-existing consequence of #777's fallback, not something this PR introduces, and repairing it means deciding what those tests should assert when reflection legitimately comes from `information_schema` — a wider change than this PR's scope.

Re-validated at `d2f9b68a`: 6 passed live.

laughingman7743: **Rebase onto `master` after #784 merged, with the upstream overlap checked.**

`7964872..e67c8db` replaces the earlier `7964872..d2f9b68`. Two hunks conflicted, both in the code #784 touched, and the resolution changed what this PR claims:

1. `pyathena/sqlalchemy/base.py` — #784 added `"string" if data_type == "varchar" else data_type` to `_columns_from_information_schema`, on the same line as this PR's comment normalization. Kept both. The consequence is that the type asymmetry this PR's live test asserted no longer exists: `information_schema` reported a Hive `STRING` as unbounded `varchar`, and the dialect now maps it back, so reflected types match the metadata API whichever path answered. The test's type assertion follows in `e67c8db` — `types.String` rather than `types.VARCHAR`, compared by exact class because `VARCHAR` is a `String` subclass.

   Worth noting for the record: that assertion used to double as proof the fallback had actually run, since only the fallback could yield `VARCHAR`. It no longer distinguishes the paths. The `StartQueryExecution` recording added in `650f38c` covers that instead, and it covers it better — it names the query rather than inferring it from a type.

2. `tests/pyathena/conftest.py` — #784 had already added `unload` to `create_engine`'s option allowlist (and to `create_async_engine`). This PR's identical one-line addition is gone; the change is upstream, not duplicated.

Nothing else in the upstream delta touches the contracts here: the binary colspecs, `AthenaBinary`, and the pandas/Arrow CSV NULL work are separate paths from comment normalization, and `docs/sqlalchemy.md` moved only in the type-name prose that #784 owns.

Re-validated at `e67c8db`: `just lint` clean, `TestAthenaDialect` 6 passed, and live `test_throttled_columns_across_cursor_types` (6 engines) plus `test_get_table_comment` — 7 passed. CI for this head is pending; the last full green was `e5001d2`, which predates the rebase and the per-case-table change.

Issue #786, opened 2026-09-20
get_table_comment and get_table_options have no throttling fallback


laughingman7743:
## Problem

`get_table_comment()` and `get_table_options()` have no fallback when Athena throttles the metadata API, so whole-table reflection fails under the conditions #777 exists to survive.

Measured against a live table while `GetTableMetadata` was refused:

```
get_table_comment : OperationalError
get_table_options : OperationalError
get_columns       : OK   (information_schema fallback)
has_table         : OK   (information_schema fallback)
```

`Table(name, metadata, autoload_with=conn)` calls all four, so it raises even though the columns themselves are reachable. Only `has_table()` and `get_columns()` are throttling-tolerant today.

## Why it is this way

#777 chose it deliberately: `information_schema` has no table comment, no `LOCATION`, no SerDe and no table properties, so it cannot answer these two methods. That reasoning holds for `information_schema`.

It did not consider `SHOW CREATE TABLE`, which returns all of it — table comment, `PARTITIONED BY`, `ROW FORMAT SERDE`, `STORED AS INPUTFORMAT`/`OUTPUTFORMAT`, `LOCATION` and `TBLPROPERTIES` — as a query rather than a metadata API call. The dialect already runs `SHOW CREATE VIEW` this way in `get_view_definition()`.

## What a repair would involve

Parsing DDL text back into the structured values `get_table_options()` returns (`awsathena_location`, `awsathena_compression`, `awsathena_row_format`, `awsathena_file_format`, `awsathena_serdeproperties`, `awsathena_tblproperties`). That is a new parser for a format Athena controls and can reformat, which is a different kind of risk from the rest of the reflection code, and it needs to agree with what the metadata API reports for the same table — including the cases where they legitimately differ.

Worth deciding before writing it:

- Is whole-table reflection under throttling a real user problem, or is the current behavior (fail loudly, retry) acceptable? #780 measured the throttling episodes but not how often applications reflect whole tables during one.
- If `SHOW CREATE TABLE` is adopted, does it replace the metadata API on the fallback path only, or is it preferable everywhere (one query instead of one API call, no metadata rate limit)?
- A query costs ~2s versus ~0.2s for the API call (#777's measurement for `information_schema`), so it is a fallback, not a default.

Alternatively, document the boundary — `has_table()` and column reflection are throttling-tolerant, whole-table reflection is not — and close this.

## Context

- #777, #780 for the throttling fallback and the measurements behind it.
[...]

PR #784, opened 2026-09-20
Preserve Hive string types in metadata fallback reflection

laughingman7743:
## WHAT

Normalize unbounded `varchar` from `information_schema.columns` to SQLAlchemy `String`, matching reflection of Hive `STRING` through Athena's metadata API.
Keep bounded `VARCHAR(n)` and `CHAR(n)` types and lengths unchanged.

Extend the existing Hive string reflection integration test to query the fallback directly as well as use the Inspector, with exact type and length assertions.
Both paths execute against Athena through the existing fixtures; no new mocks are used.
Update the existing unit expectation and the fallback documentation.

## WHY

Follow-up to #780 and the information_schema fallback introduced in #777.
This fixes the string-type mismatch on that fallback path.

After a throttled metadata request, the fallback returned `VARCHAR()` where the metadata API returned `String()`.
This made reflected types depend on throttling and failed the existing `test_create_table_with_varchar_text_column` on Python 3.11 and 3.12 in [PR #770's CI](https://github.com/pyathena-dev/PyAthena/actions/runs/35499655423).
PR #770 has since merged after a successful CI rerun; this follow-up fixes the metadata type discrepancy observed in its earlier run.

Validation:

- Before the implementation fix, the expanded live test failed only for the fallback's unbounded string case: 1 failed, 5 passed.
- After the fix, the same six live cases passed with REST and all six passed with async REST, run sequentially with `-n 1`.
- The existing failing VARCHAR/Text reflection integration test and the existing fallback conversion unit test both passed (2 tests).
- `just format`, `just lint`, and `just docs lint` passed.
- The current-head CI passed all 18 checks, including the five Python versions for regular, synchronous SQLAlchemy and asynchronous SQLAlchemy suites, docs lint/build, and benchmark tooling.

- Both self-review rounds and the independent Claude Code static review returned CLEAN; records are inline.
- [Current-head CI](https://github.com/pyathena-dev/PyAthena/actions/runs/35500855810) passed. The post-#770 master integration was also checked for conflicts and preservation of the four-file patch.

PR #798, opened 2026-09-23
Read the dialect's own queries through an API cursor and resolve federated absence from information_schema

### LINKED ISSUE #781: Verify metadata error handling and the information_schema fallback on federated catalogs and DataFrame cursors

### LINKED ISSUE #783: Verify metadata error handling and the information_schema fallback on federated catalogs
## Problem

PR #777 (#780) changed how SQLAlchemy reflection distinguishes "table does not exist" from "the metadata request failed", and added an `information_schema.columns` fallback for `has_table()` and column reflection when Athena throttles `GetTableMetadata`.
Both behaviors were verified against the Glue Data Catalog (`AwsDataCatalog`) and, for listing semantics, an S3 Tables catalog.
#782 added live coverage through the DataFrame/UNLOAD cursor classes.
Neither covers Lambda-based or Hive-metastore federated catalogs, and the test account has none.

This issue carries over items 1-3 of #781; item 4 shipped in #782.

## What is unverified

1. **Missing-table detection in federated catalogs.** `pyathena/sqlalchemy/base.py` maps a `MetadataException` to `NoSuchTableError` only when its message carries the `(Service: AmazonDataCatalog; ...; Error Code: EntityNotFoundException; ...)` envelope (`pyathena/util.py`, `_get_error_code(..., unwrap_metadata=True)`). Before #777 any `MetadataException` meant "missing". If a federated catalog reports a missing table with a different message, `has_table()` now raises `OperationalError` instead of returning `False`, and `create_all(checkfirst=True)` fails.
2. **`information_schema.columns` in federated catalogs.** The fallback assumes the query is served for the cursor's catalog, that names are reported in lowercase, and that `extra_info = 'partition key'` marks partition columns.
3. **`ListTableMetadata` `Expression` for names over 128 characters.** `_lookup_table` lists with `re.escape(name.lower())` and matches case-insensitively, but `_fold_table_name` only lowercases for `awsdatacatalog`, so outside Glue the cache key keeps the caller's casing while the server-side filter is forced to lowercase. The filter semantics of non-Glue catalogs are unknown.

## Infrastructure this needs

The test account has no federated connector. The S3 Tables setup is the precedent to follow:

- `scripts/cloudformation/github_actions_oidc.yaml` declares the catalog resource (`AWS::Glue::Catalog` with `FederatedCatalog`, lines 362-386) — a Lambda connector would instead need `AWS::Athena::DataCatalog` with `Type: LAMBDA`, plus the connector's own SAR deployment. The repo has no SAM or `AWS::Serverless::*` precedent, and `test-suite.yaml` already documents "provisioned out of band" as the accepted escape hatch for resources the template cannot express.
- IAM additions: `lambda:InvokeFunction` on the connector function. `athena:*` already covers `CreateDataCatalog`/`GetDataCatalog`, and a spill prefix inside the existing staging bucket is already covered by `s3-access`.
- `AWS_ATHENA_FEDERATED_CATALOG` in `tests/__init__.py` and `.github/workflows/test-suite.yaml`, with a `requires_federated_catalog` skipif mirroring `requires_s3_tables` (`tests/pyathena/sqlalchemy/test_base.py`).
- Test-side cost is otherwise low: `tests/pyathena/conftest.py::create_engine` already accepts `catalog_name`, so `{"catalog_name": ENV.federated_catalog}` with `indirect=True` is all a test needs.

## Proposed verification

Record, for a missing table in a federated catalog: the raw `MetadataException` message, whether `information_schema.columns` returns rows with the expected `extra_info`, and how `ListTableMetadata` treats `Expression`.
If the missing-table envelope differs, decide whether to recognize the additional format explicitly rather than restoring the blanket `MetadataException` mapping, which turned throttling into false missing-table results.

`docs/sqlalchemy.md` documents the current behavior in Glue terms; update it once federated catalogs are verified.

## Context

- #781 (items 1-3, carried here), #782 (item 4, DataFrame/UNLOAD cursors), #777 merged as `7964872`, #780.

### LINKED ISSUE #785: get_view_definition returns a corrupted or failing result on DataFrame dialects
## Problem

`AthenaDialect.get_view_definition()` returns a corrupted view definition, or raises `TypeError`, on every DataFrame-backed dialect when the view's DDL contains a blank line.

The dialect runs `SHOW CREATE VIEW` through `connection.scalars(text(query))` and joins the rows with `"\n".join(res)` (`pyathena/sqlalchemy/base.py`). That query is executed by whichever cursor class the user configured, and those cursors disagree on how a blank row materializes.

Measured against a view whose formatted DDL has one blank line:

| dialect | result |
|---|---|
| `awsathena+rest` | OK — 16 lines, blank line at index 6 |
| `awsathena+pandas` | `TypeError: sequence item 6: expected str instance, float found` |
| `awsathena+arrow` | **silently wrong** — 15 lines, the blank line is gone |
| `awsathena+polars` | `TypeError: sequence item 6: expected str instance, NoneType found` |

Reproduce with any view Athena formats with a blank line, for example:

```sql
CREATE OR REPLACE VIEW v AS
WITH t AS (SELECT 1 AS a, 'x' AS b)
SELECT a, b FROM t UNION ALL SELECT 2, 'y'
```

then `sqlalchemy.inspect(conn).get_view_definition("v")` on each dialect.

The arrow case is the worst of the three: it neither raises nor round-trips, so a reflected view definition differs from the one Athena reported.

## Cause

This is the same root as the comment defect repaired in #782: the dialect parses result rows from queries it issues itself, but those queries inherit the user's result format. `pyathena/arrow/result_set.py` reads a CSV NULL as `''`, `pyathena/pandas/result_set.py` reads it as `NaN`, and the UNLOAD/Parquet paths read it as `None` — documented, intentional cursor behavior (`docs/null_handling.md`), not a cursor bug.

#782 handled it by normalizing values where they enter `_columns_from_information_schema`. That works for one known column but does not generalize: for `SHOW CREATE VIEW` the arrow path loses the row entirely, so no value-level normalization can recover it.

The dialect issues exactly two queries whose rows it parses itself:

- `_columns_from_information_schema` — the throttling fallback added in #777
- `get_view_definition` — `SHOW CREATE VIEW`

## Proposed fix

Run both through the plain API `Cursor` rather than the connection's configured cursor class, so dialect internals do not inherit a result format chosen for user queries.

`Connection.cursor(Cursor)` already selects the class (`_cursor = cursor or self.cursor_class`). Checked while diagnosing:

- `BaseCursor.get_default_converter()` returns `DefaultTypeConverter()` regardless of `unload`, and `Cursor` does not override it, so the converter is right as long as the connection has no user-supplied `converter`.
- `Connection.cursor()` applies `kwargs.update(self.cursor_kwargs)` first, so a configured `unload=True` still reaches the call. It is harmless here: `Cursor` has no `_unload` attribute, so `_prepare_unload` does not wrap the query, and `BaseCursor.__init__` absorbs the leftover keyword.

Consequences to weigh:

- Reflection stops honoring a `cursor_class` set for instrumentation. This changes the shape #777 shipped.
- The `isinstance(comment, str)` guard in `_columns_from_information_schema` becomes unnecessary.
- Under `unload=true`, the fallback stops writing a Parquet UNLOAD to S3 for a five-row metadata lookup.
- A connection-level `converter=` would still reach the internal cursor; decide whether to override it explicitly.

## Coverage

`get_view_definition` has no test that exercises a view whose DDL contains a blank line, and none that runs it on a DataFrame dialect. The fix needs both.

## Context

- #782 repaired the comment half of this and added `test_throttled_columns_across_cursor_types`, whose execution-mode assertion will need to invert: the fallback would no longer run as UNLOAD for any engine.
- #777, #780 for the throttling fallback this sits on.

laughingman7743:
## WHAT

Two reflection defects with one cause: the SQLAlchemy dialect parses the rows of queries it issues itself, and it ran those queries through whatever cursor class the connection was configured with, or decided absence from an error message it could not read.

**`get_view_definition()` on DataFrame dialects (#785).** `SHOW CREATE VIEW` returns the definition one row per line. Measured against a view whose formatted DDL contains a blank line:

| dialect | before |
|---|---|
| `awsathena+rest` | correct, 16 lines |
| `awsathena+pandas` | `TypeError: sequence item 6: expected str instance, float found` |
| `awsathena+arrow` | 15 lines — the blank line silently dropped |
| `awsathena+polars` | `TypeError: sequence item 6: expected str instance, NoneType found` |

The dialect's own queries — `SHOW CREATE VIEW` and the `information_schema.columns` fallback — now open a plain API cursor through `_internal_cursor()`, with the default converter pinned, whatever `cursor_class` or `unload` the connection carries. The async adapter accepts a cursor class and maps `Cursor` to `AioCursor`. Under `unload=true` the fallback also stops writing a Parquet UNLOAD to S3 for a five-row metadata lookup.

A missing view is a query Athena runs and fails, raised without an underlying API error, and only that maps to `NoSuchTableError`. A failed `GetQueryResults` call propagates, including the first page that `execute()` prefetches. So does a `DatabaseError` from a failed `StartQueryExecution`.

**`has_table()` on Lambda federated catalogs (#781 / #783 item 1).** Measured against a live `AthenaDynamoDBConnector` catalog, a missing table comes back as:

```
MetadataException: Failed to invoke lambda function due to ... Requested resource not found
(Service: DynamoDb, Status Code: 400, Request ID: ...)
```

There is no `(Service: AmazonDataCatalog; ...; Error Code: EntityNotFoundException; ...)` envelope, so since #777 `has_table()` raised `OperationalError` instead of returning `False`, and `create_all(checkfirst=True)` failed against such a catalog. The wording is the connector's, so recognizing this string would not generalize.

A `MetadataException` that survives unwrapping now goes to the existing `information_schema` fallback, **but only outside `AwsDataCatalog`**. The fallback runs against the same catalog through the same connector: 0 rows means absent, a failing connector fails the query too, so an outage is not reported as absence.

Glue is excluded on purpose. It reports missing tables and permission failures in a recognized envelope, so an unrecognized message there has an unknown cause. `information_schema` filters by Lake Formation instead of failing, so answering from it would report a table the caller cannot see as absent — the false absence #777 exists to prevent. `_is_fallback_error()` holds that boundary. An S3 Tables catalog is not `AwsDataCatalog`, but it was measured to return the `AmazonDataCatalog` / `EntityNotFoundException` envelope, so `_lookup_table` maps it before the fallback is reached and its behavior is unchanged.

## WHY

#781 asked whether #777's error handling and fallback hold on federated catalogs. Verified live against a temporary DynamoDB connector catalog; results are recorded on #781:

- **Item 1 — missing-table detection:** broken as suspected. Repaired here.
- **Item 2 — `information_schema.columns` in a federated catalog:** works. Same type names as `GetTableMetadata`, lowercase names, and **0 rows rather than an error** for a missing table or schema. That last property is what makes the item 1 repair sound.
- **Item 3 — `ListTableMetadata` `Expression` over 128 characters:** works. A 140-character name matched with the lowercase regex `_lookup_table` sends, the original case, and a prefix regex.

`get_view_definition()` was found while comparing both reflection paths after #784 and shares the root cause of the arrow comment defect repaired in #782. #782 normalized values at one column; `SHOW CREATE VIEW` loses whole rows on arrow, so no value-level normalization could recover them.

## Behavior changes

- Reflection no longer runs through a `cursor_class` set for instrumentation. The dialect's own queries always use the API cursor.
- `Inspector.get_view_definition()` can now raise a raw `pyathena.error.*` instead of a SQLAlchemy-wrapped `DBAPIError`, because it no longer goes through `Connection.scalars()`. The `information_schema` fallback has behaved this way since #777.
- In a non-Glue catalog, an unrecognized `MetadataException` from `GetTableMetadata` is no longer retried. `information_schema` is asked immediately instead, as for throttling. `_without_throttling_retries` is renamed `_without_fallback_retries`; the excluded code set is unchanged.
- The compliance-suite case `test_metadata_errors_do_not_establish_absence[None]` now uses a permission message. It still asserts propagation, which is correct because that suite runs against Glue. The federated case has no catalog in CI and is covered by `TestAthenaDialect::test_unrecognized_metadata_error_asks_information_schema`.

## Validation

- `just format`, `just lint` (ruff, mypy, cfn-lint), `just docs lint` pass.
- `TestAthenaDialect` (no AWS): 11 passed.
- Local live runs: `tests/pyathena/sqlalchemy/` + `tests/pyathena/aio/sqlalchemy/` — 446 passed at the rebased head. `just test sqla` — 510 passed; `just test sqla-async` — 509 passed. Both suite runs were before the last review repair and rebase, which CI covers.
- Regression coverage fails without the fix. Reverting `_internal_cursor()` to the user's cursor fails `test_get_view_definition_across_cursor_types` for pandas, pandas+unload and arrow. That test compares the full reflected definition with the rows Athena returned through an API cursor, and skips only if Athena formats the view without a blank line.
- Federated catalog, live: `has_table('this_table_does_not_exist')` changed from `OperationalError` to `False`, `has_table` on an existing table is `True`, and `get_columns` on the 140-character name returns the right columns.
- CI at `8e90677` passed every job: lint, docs build, offline, and 5 each of `test`, `test-sqla` and `test-sqla-async`.
- Reviews: `/code-review` (7 findings), a claims audit, an independent Codex review (3 findings), and two independent follow-ups. Every actionable finding was repaired, including two regressions this branch had introduced: result-paging failures reported as a missing view, and the comment guard dropped although a `cursor_kwargs` converter can still produce NaN. Records follow inline.

## Limits

- **Still unverified on federated catalogs:** `extra_info = 'partition key'`, because a DynamoDB backend has no partition columns, and case-preserving connectors, because the DynamoDB connector lowercases names in both `ListTableMetadata` and `information_schema`. Both need a Hive or similar connector.
- A `converter` given in `cursor_kwargs` still reaches the internal cursor, because `Connection.cursor()` applies `cursor_kwargs` last. The `isinstance(comment, str)` guard keeps that from producing NaN comments.
- `get_view_definition()` still reports a CANCELLED query, or one FAILED for a reason other than a missing view, as absence. This predates the branch: master mapped every `OperationalError` to `NoSuchTableError`, and this branch removes only API failures from that. The cursor does not keep the failed query's state, and a missing view reports `ErrorCategory 1` / `ErrorType 1502`, which does not single it out.
- `get_table_comment()` and `get_table_options()` still have no throttling fallback. Tracked in #786.
- The temporary connector stack, Athena data catalog and DynamoDB tables are deleted after this merges.

Closes #781, closes #783, closes #785. Refs #777, #782, #784, #786


pyathena/sqlalchemy/base.py
```diff
@@ -380,21 +385,58 @@ def _get_columns(self, connection, table_name: str, schema: str | None = None, *
         return self._columns_from_metadata(metadata)
 
     @staticmethod
-    def _without_throttling_retries(retry_config: RetryConfig) -> RetryConfig:
-        """Copy a policy without the codes that carry throttling.
+    def _answerable_from_information_schema(code: str | None, catalog: str | None) -> bool:
```
laughingman7743: **Self-review round one — implementation behavior** (`/code-review`, high). Base `15325e172714e2e809dd5c597e8babb2c51a6fac` (rebased; reviewed originally at `d7e1dcd..580cf7a`), head `205c9b14d4e152b39c810609455852f22d38512d`. Surfaces: `_get_columns` fallback eligibility, `_internal_cursor`, `_columns_from_information_schema`, `get_view_definition`, the async adapter, the retry-policy derivation, and both test trees.

Result: **FINDINGS**, repaired.

1. **High, repaired in `82d5648`.** `MetadataException` in the fallback set turned unparsed permission failures into false absence. `_get_error_code` unwraps only the `AmazonDataCatalog` envelope, and `information_schema` filters by Lake Formation instead of erroring, so it returns 0 rows for a table the caller cannot see. The fallback is now gated on the catalog not being `AwsDataCatalog`, in `_answerable_from_information_schema` (line 388).
2. **Medium, disproven by measurement.** "The fallback lowercases, so a mixed-case federated table is missed." The DynamoDB connector reports `PyAthenaFederatedProbeMixedCase` as `pyathenafederatedprobemixedcase` in both `ListTableMetadata` and `information_schema`. No change. A case-preserving connector remains unverified (see PR Limits).
3. **Medium, repaired by test.** A failed `StartQueryExecution` raises `DatabaseError`, the parent of `OperationalError`, so it is not caught. That is correct: a request that never ran is not a missing view. `test_get_view_definition_propagates_a_failed_request` pins it.
4. **Low, repaired in `82d5648`.** A connection-level `converter` would reach the internal cursor, so the default converter is now pinned. The follow-up is under the independent review below.
5. **Low, repaired in `a3fbf4d`.** The propagation case for an unrecognized Glue error had been removed. It is restored with a permission message.
6. **Low, repaired, then superseded by the independent review.** The blank-line assertion depended on Athena's formatter.

Also found myself: the async adapter did not accept a positional cursor class, and could not drive a synchronous `Cursor` anyway. `just test sqla-async` failed 3 cases on it. Repaired in `c61c361` by mapping `Cursor` to `AioCursor`.

docs/sqlalchemy.md
```diff
@@ -100,12 +100,16 @@ A later listing preserves metadata already fetched for a table.
...
-A `retry_config` in `cursor_kwargs` replaces that policy entirely, including its throttling retries, which then run before the fallback.
+Column reflection and `has_table()` do not retry a table-metadata request that `information_schema` can answer; they read `information_schema.columns` instead, executed without query result reuse, and log a warning.
+That covers a throttled request in any catalog.
+It also covers a `MetadataException` carrying no recognized Glue error envelope, but only outside `AwsDataCatalog`: a federated catalog reports a missing table in its connector's own words, so absence is decided by the query against that catalog rather than by an unrecognized message.
+In `AwsDataCatalog` an unrecognized `MetadataException` still propagates, because Glue does state missing tables and permission failures in a recognized envelope, and `information_schema` filters by Lake Formation instead of failing, so reading it there would report a table the caller cannot see as absent.
```
laughingman7743, line 106: **Self-review round two — claims and operational behavior.** Base `15325e172714e2e809dd5c597e8babb2c51a6fac`, head `205c9b14d4e152b39c810609455852f22d38512d`. Claims audited: this docs paragraph, the docstrings of `_internal_cursor` and `_answerable_from_information_schema`, the commit messages, and the PR body.

Result: **FINDINGS**, corrected.

1. The docs still said any `MetadataException` without an envelope goes to the fallback, after the code had been narrowed to non-Glue catalogs. Corrected in `8ecd6bf`. The paragraph now states the catalog boundary and the Lake Formation reason for it.
2. The claim "only outside `AwsDataCatalog`" was not yet checked against S3 Tables, which is outside `AwsDataCatalog` but is Glue-federated. Measured live: a missing S3 Tables table returns `Entity Not Found (Service: AmazonDataCatalog; ...; Error Code: EntityNotFoundException; ...)`, so `_lookup_table` maps it before the fallback. S3 Tables behavior is unchanged. The PR body says so.
3. `get_view_definition` against a missing view was measured live on rest and pandas: `NoSuchTableError` on both. The stub test for that path cites this.

Operator view: the fallback adds one `information_schema` query per unrecognized federated failure, instead of that failure's retry ladder. Internal queries no longer UNLOAD, which removes an S3 write and adds no query.

Deferred and stated in the PR: `extra_info` partition marking and case-preserving connectors on federated catalogs remain unverified. The raw `pyathena.error.*` from `get_view_definition` is not SQLAlchemy-wrapped. A `cursor_kwargs` converter still reaches the internal cursor.

pyathena/sqlalchemy/base.py
```diff
@@ -523,12 +566,17 @@ def get_view_definition(
...
-        else:
-            return "\n".join(res)
+        with self._internal_cursor(raw_connection) as cursor:
+            try:
+                cursor.execute(query)
+            except pyathena.error.OperationalError as e:
+                # Only a rejected query says the view is absent. A failure while
+                # paging the results is a failed read of a view that does exist.
+                raise exc.NoSuchTableError(f"{schema}.{view_name}") from e
+            rows = cursor.fetchall()
```
laughingman7743, line 587: **Independent review — relayed result.** Reviewer: Codex, a different model from the author, through the repository's Codex rescue agent. Read-only static inspection, with no edits, tests or GitHub access. Reviewed `d7e1dcd..5f87ed3`; the rebased equivalent of that head is `8ecd6bf`. Surfaces the reviewer read: all changed files, plus `util.py`, `connection.py`, `common.py`, `cursor.py`, `converter.py` and `result_set.py`, the async connection/cursor/adapter paths, and the pandas/arrow/polars cursor and result-set paths.

Result: **FINDINGS**, 3 items, all repaired in `205c9b1`.

1. **P2.** `fetchall()` sat inside the `OperationalError -> NoSuchTableError` handler. A `GetQueryResults` page that exhausted its retries (`result_set.py:370`) reported an existing view as absent. This was a regression introduced by this branch: before it, iteration was outside the handler. Now only `execute()` is inside the handler.
2. **P2.** `Connection.cursor()` applies `cursor_kwargs` after the explicit arguments, so a `cursor_kwargs["converter"]` still overrides the pinned one. Dropping the `isinstance(comment, str)` guard had therefore reopened the NaN-comment path. The guard is restored at line 482, and the stub test has its NaN row back.
3. **P2.** The blank-line test could skip on the defect itself. Losing every blank row left `CREATE VIEW` and `UNION ALL`, so both assertions passed and the skip blamed Athena's formatting. The test now compares the full definition against rows fetched through an API cursor (line 949), and skips only when those rows have no blank line.

A follow-up on `8ecd6bf..205c9b1` has been requested and will be recorded in reply.

laughingman7743: **Independent follow-ups on the repairs** (same Codex reviewer, read-only, static inspection).

Follow-up 1, `8ecd6bf..205c9b1`: item 2 (comment guard) **CONFIRMED-RESOLVED**. Two new **FINDINGS**, both repaired in `185f582`:

- `Cursor.execute()` prefetches the first result page (`result_set.py` `_pre_fetch`), so narrowing the handler to `execute()` still caught a failed first `GetQueryResults` as a missing view. Measured live: a missing view is a query Athena runs and fails (`View not found or not a valid presto view: ...`), raised as `OperationalError` with `__cause__ is None`. An API failure is raised `from` its `ClientError`. Only the cause-less form maps to `NoSuchTableError` now. My earlier missing-view stub had modelled a rejected `StartQueryExecution`, which actually raises `DatabaseError` and never reaches this handler. It is replaced, and a failed-first-page test is added.
- The live test's baseline is read through the same API-cursor path as the fix, so a loss on both sides would go unnoticed. The independent `CREATE VIEW` / `UNION ALL` content checks are restored alongside the full comparison.

Follow-up 2, `205c9b1..185f582`: stubs and live test **CONFIRMED-RESOLVED**. One remaining **FINDING**, recorded rather than fixed in `bde5075`: a CANCELLED query, or one FAILED for a reason other than a missing view, also raises a cause-less `OperationalError` and is still reported as absence. The reviewer confirmed this mapping predates the branch. Master mapped every `OperationalError` to `NoSuchTableError`; this branch only removed API failures from it. Distinguishing further would need a `GetQueryExecution` call on the failure path. I measured one: a missing view reports `ErrorCategory 1` (SYSTEM) and `ErrorType 1502`, which does not single it out, so keying on it would be a guess. The comment now states the limit instead of implying a precise test, and the PR body lists it. No has-cause-but-failed-query path was found, sync or async.

Live after the repairs: a missing view gives `NoSuchTableError` on rest and pandas, and `test_get_view_definition_across_cursor_types` passes 5 of 5.

pyathena/sqlalchemy/base.py
laughingman7743: Renamed `_answerable_from_information_schema` to `_is_fallback_error` in `8e90677`, to match `_FALLBACK_ERROR_CODES` and `_without_fallback_retries`. The docstring now also says the catalog, not only the code, decides the result. No behavior change.

pyathena/sqlalchemy/base.py
laughingman7743: **Final independent follow-up**, `185f582..8e90677` (same Codex reviewer, read-only): **CONFIRMED-RESOLVED**. There is no behavior change. The only edits are the rename, the docstring, and the new limit comment in `get_view_definition()`, and the decision logic of `_is_fallback_error` is unchanged. `git grep` finds no stale reference to the old name in tracked code, tests or docs. The new comment and docstring match the code: a failed or cancelled query raises a cause-less `OperationalError` and is read as absence, while API-layer failures keep their cause and propagate. The ErrorCategory/ErrorType measurement was taken as given, not re-derived.

PR #800, opened 2026-09-23
Find an existing table in to_sql whatever the name's case

### LINKED ISSUE #799: to_sql misses an existing table when the name has uppercase letters
## Problem

`pyathena.pandas.util.to_sql()` does not find an existing table whose name was given with uppercase letters, so `if_exists="fail"` does not fail and data is appended twice.

`to_sql` checks existence with:

```sql
SELECT table_name FROM information_schema.tables
WHERE table_schema = '{schema}' AND table_name = '{name}'
```

It passes `schema` and `name` through unchanged. Athena stores names in lowercase and reports them in lowercase in `information_schema`, so `name="MyTable"` never matches the existing `mytable`.

Reproduced against live Athena on `master` (`3c69daf`):

```python
to_sql(df, "PyAthenaToSqlCase…", conn, location, schema="default", if_exists="fail")  # creates the table
to_sql(df, "PyAthenaToSqlCase…", conn, location, schema="default", if_exists="fail")  # no error
# SELECT count(*) -> 2   (one row written twice)
```

- `if_exists="fail"` writes Parquet into the existing table's location and runs `CREATE EXTERNAL TABLE IF NOT EXISTS`, which is a no-op, so the rows are appended instead of raising.
- `if_exists="replace"` follows the same path: the existing table is not dropped and its S3 objects are not deleted, so it appends too. This follows from the code and was not reproduced separately.

The existing `test_to_sql` uses a lowercase generated name, so it cannot see this.

## Same query, same defect class

- The check runs through `conn.cursor()` without disabling query result reuse. A connection with `result_reuse_enable=True` can get a cached "no rows" answer from before the table existed, and the existence check fails the same way. #777 turned result reuse off for the dialect's `information_schema` fallback for this reason.
- `schema` and `name` are interpolated into string literals without escaping `'`.

## Proposed fix

Lowercase and quote-escape both identifiers in the existence check, the way `_columns_from_information_schema` does in the SQLAlchemy dialect. Execute the check with `result_reuse_enable=False`. Add a regression test that runs `to_sql` twice with a mixed-case name: once expecting `if_exists="fail"` to raise, and once expecting `if_exists="replace"` to leave one copy of the rows.

## Out of scope

- In `AwsDataCatalog`, `information_schema` filters by Lake Formation rather than failing, so a table the caller cannot see is not found either. This is the same limit #798 documented for the dialect.
- The DDL, `DROP TABLE` and `ALTER TABLE` statements quote identifiers with backticks. Athena resolves those case-insensitively, so they are unaffected.

laughingman7743:
## WHAT

`pyathena.pandas.util.to_sql()` now finds an existing table whose name contains uppercase letters.

The existence check compared `schema` and `name` verbatim against `information_schema.tables`, but Athena stores identifiers in lowercase and reports them that way. It now compares lowercase, quote-escaped literals, the same way the SQLAlchemy dialect builds its `information_schema` fallback query.

`test_to_sql` now uses a mixed-case table name (`To_Sql_<uuid>`). Its existing sequence therefore pins the existence check where it matters: the second `if_exists="fail"` must raise, and `replace` must leave exactly one row. `append` does not consult the check, so that step runs with the mixed-case name without testing the check.

## WHY

With a mixed-case name the check never matched (#799). Reproduced on `master`:

```
first : created
second: NO ERROR (existing table not detected)   # if_exists="fail"
rows  : 2                                        # the row was written twice
```

`if_exists="fail"` wrote Parquet into the existing table's location, and `CREATE EXTERNAL TABLE IF NOT EXISTS` was then a no-op, so the call silently appended instead of raising. `if_exists="replace"` takes the same branch, so by the code it appends instead of dropping and rewriting; that was not reproduced separately on `master`, but the new test's replace step asserts the corrected outcome. The documented example in `docs/pandas.md`, `to_sql(df, "YOUR_TABLE", ..., if_exists="replace")`, is exactly that case. `test_to_sql` used to generate a lowercase name, so it could not see the defect.

The DDL, `DROP TABLE` and `ALTER TABLE ... ADD PARTITION` statements quote identifiers with backticks. Athena resolves those case-insensitively, so they needed no change.

## Validation

- `just format`, `just lint` pass.
- Live, `-n 1`: `test_to_sql` passes, and all of `tests/pyathena/pandas/test_util.py` passed before the fold (12 passed).
- `test_to_sql` fails without the fix. With the pre-fix `util.py` checked out it reports `DID NOT RAISE OperationalError` on the second `if_exists="fail"` call.
- CI at `5e3137c` passed every job. One `test (3.13)` job first failed in the setup of the unrelated `TestAioCursor.test_executemany_cancellation`, with `OperationalError: Invalid S3 request` from an aio cursor fixture. Rerunning the failed job passed.
- `/simplify` folded a separate mixed-case test into `test_to_sql`. It repeated the same sequence, and folding it removes seven billed queries and one leftover table per Python version.
- #799 also raised a concern about query result reuse, which was checked and dropped. In a workgroup where reuse works, the existing reuse test's query is reused on its second run, but an `information_schema.tables` query run three times was never reused. So no cached "absent" can answer this check. The guard was removed: its test case passed with and without the change.

Not exercised by the test: a mixed-case `schema`, partitions, and names containing `'`. For lowercase names without `'`, the generated query is unchanged.

Limits: in `AwsDataCatalog`, `information_schema` filters by Lake Formation instead of failing. A table the caller cannot see is therefore still not found, the same limit documented for the dialect in #798.

Closes #799


pyathena/pandas/util.py
```diff
@@ -261,13 +261,17 @@ def to_sql(
     ).Bucket(bucket_name)
     cursor = conn.cursor()
 
+    # Athena stores identifiers in lowercase and information_schema reports them
+    # that way, so compare lowercase literals.
+    schema_literal = schema.lower().replace("'", "''")
```
laughingman7743: **Self-review round one — implementation behavior** (`/code-review`, high). Base `3c69daf44feb7704f96e1e6664681a8ddabf2791`, head `3853502`, re-checked at `5e3137cdd860fd9107c15c85a863f7b7ba87ee63`. Surfaces covered: the existence check, the `fail`/`replace`/`append` branches, DDL/`DROP`/`ALTER` identifier handling, and the S3 deletion on replace.

Result: **CLEAN** for the fix. Doubling `'` is sufficient in Athena string literals, and since the check is executed without parameters, a `%` in a name cannot break it. The later statements quote identifiers with backticks, which Athena resolves case-insensitively. This is the only existence check of its kind outside the dialect.

A claim of my own was also corrected in this round. #799 said query result reuse could answer the check with a stale "absent". I measured it in `ENV.work_group`, where reuse works: the existing reuse test's query was reused on its second run, but an `information_schema.tables` query was never reused in three runs. The guard added for it in `3853502` was removed in `0a4eb50`, because its test case passed both with and without the change.

tests/pyathena/pandas/test_util.py
```diff
@@ -342,7 +342,9 @@ def test_to_sql(cursor):
             "col_binary",
         ]
     ]
-    table_name = f"""to_sql_{str(uuid.uuid4()).replace("-", "")}"""
+    # Uppercase letters: Athena reports names in lowercase, and the existence
+    # check behind if_exists has to find the table anyway.
+    table_name = f"To_Sql_{uuid.uuid4().hex}"
```
laughingman7743: **Self-review round two — claims and quality** (`/simplify`: reuse, simplification, efficiency, altitude). Base `3c69daf44feb7704f96e1e6664681a8ddabf2791`, head `5e3137cdd860fd9107c15c85a863f7b7ba87ee63`.

- **Efficiency, repaired in `5e3137c`.** The separate mixed-case test repeated `test_to_sql`'s fail/fail/replace sequence, at the cost of 7 billed queries and one leftover table per Python version. It is folded into `test_to_sql` by giving that test a mixed-case name. With the pre-fix `util.py`, `test_to_sql` fails with `DID NOT RAISE OperationalError`.
- **Altitude: no change.** The reviewers considered normalizing `name` once at the top, calling `GetTableMetadata`, and reusing the dialect's `has_table`. Each would widen the blast radius or reintroduce the metadata throttling from #777. The literal-level fix matches `_columns_from_information_schema`.
- **Reuse: skipped.** Parameter binding through the formatter, or a shared lowercase-and-escape helper for the two call sites, would be marginal and would touch the dialect. The current two lines match the formatter's Trino escaping rule.
- **Claims:** the PR body's `replace` statement is now scoped to what was reproduced. The code implies that `replace` also appended, but only `fail` was reproduced on `master`.

pyathena/pandas/util.py
```diff
@@ -261,13 +261,17 @@ def to_sql(
...
+    name_literal = name.lower().replace("'", "''")
```
laughingman7743: **Independent review — relayed result.** Reviewer: Codex, a different model from the author. Read-only static inspection with no tests or AWS access. Reviewed `3c69daf44feb7704f96e1e6664681a8ddabf2791..5e3137cdd860fd9107c15c85a863f7b7ba87ee63`. Surfaces: the lookup and its escaping; `fail`/`replace`/`append`; `DROP`/`CREATE` DDL; S3 deletion; the partition `ALTER`; the changed test.

Result: **FINDINGS**. No correctness defect was found in the implementation. One claim was overstated.

- **Test coverage, claim corrected.** `append` does not consult the existence check, so the mixed-case `append` step in `test_to_sql` cannot detect a broken check. My commit message for `5e3137c` said the fold pins the check "on every path, including append". That is wrong. The test pins `fail`, where the second call must raise, and `replace`, where exactly one row must remain. The PR body now says so. It also lists what the test does not exercise: a mixed-case `schema`, partitions, and names containing `'`. For lowercase names without `'`, the generated query is unchanged.

The first attempt at this review returned without a written verdict and was re-run with an explicit instruction to write one.

PR #801, opened 2026-09-23
Report a missing federated table from table comment and option reflection

laughingman7743:
## WHAT

`get_table_comment()` and `get_table_options()` now raise `NoSuchTableError` for a missing table in a federated (non-Glue) catalog, instead of `OperationalError`.

When the table-metadata lookup fails with a `MetadataException` that carries no recognized Glue envelope, and the catalog is not `AwsDataCatalog`, `_get_table()` asks `information_schema.columns` whether the table exists:

- no row: `NoSuchTableError`, chained to the original error;
- rows: the original `OperationalError` propagates;
- columns already reflected from `information_schema` in the same Inspector count as rows, so `Table(..., autoload_with=...)` does not repeat the query;
- the failed call caches nothing, so a transient connector error does not switch later column reflection to `information_schema`.

The catalog boundary moves out of `_is_fallback_error()` into `_is_connector_error()`, which column reflection and this path share; `_is_fallback_error()` is now throttling or that predicate.

Throttling, and every error in `AwsDataCatalog`, propagate as before.

`docs/sqlalchemy.md` now states that an error response establishes absence only as a recognized `EntityNotFoundException`, while an `information_schema` query can establish it from its result. Three older sentences in that paragraph said throttling and permission failures always propagate. That has not been true since #777 and #798, so they are corrected here.

## WHY

B-1 in #786 (handoff comment after #798). A Lambda federated catalog reports a missing table in its connector's own words. #798 measured it as `MetadataException: Failed to invoke lambda function due to ... Requested resource not found (Service: DynamoDb, ...)`. `get_columns()` and `has_table()` already answer that through `information_schema` (#798), but the two table-level methods only recognized `EntityNotFoundException`, so the error contract depended on the method.

The boundary is the one `_is_fallback_error` already applies. In `AwsDataCatalog`, `information_schema` filters by Lake Formation and returns 0 rows instead of failing, so absence is never decided from it there. Throttling is excluded here because `information_schema` can supply neither a comment nor options, so it could only turn a throttled lookup of an existing table into a slower failure.

Throttling tolerance for these two methods is the rest of #786 and will come separately. Measurements on 2026-09-23 found that `glue:GetTable` reproduces `GetTableMetadata`'s parameters exactly, for Hive, partitioned, Iceberg, and S3 Tables tables. They also found that Glue calls kept succeeding while Athena's metadata API was throttled.

## TEST

Tested commit: `c1dc23d3` (code and tests). The later commits up to `62815d1e` change only `docs/sqlalchemy.md`.

- `just lint`: passed. `just docs lint`: 0 errors on `62815d1e`.
- `uv run --env-file .env pytest -n 1 tests/pyathena/sqlalchemy/test_base.py -k "table_level_reflection or unrecognized_metadata or get_table_matches or empty_information_schema or cursor_schema_applies or without_fallback"`: 15 passed.
- Vacuity, each against this head:
  - `pyathena/sqlalchemy/base.py` restored from `origin/master`: the 4 federated cases (absent/present × comment/options) fail.
  - `_get_table()` using `_is_fallback_error()` instead of `_is_connector_error()`: both throttled cases fail.
  - Cached-columns check removed: `test_table_level_reflection_reuses_reflected_columns` fails.
- Not run live: a Lambda federated catalog. The #798 probe catalog is deleted; the tests use the connector error measured there.
- CI on the first head (`36c8126d`, run 35864833745): `test / run (3.11)` and `(3.12)` failed in `tests/pyathena/pandas/test_async_cursor.py`, where `AsyncPandasCursor` returned an empty result (`test_arraysize`: `fetchmany()` gave `[]`, and `test_description`: `fetchall()` gave `[]`). The two failures are different tests on different Python versions, and this PR does not touch the pandas cursor path. The root cause is not established. `test-sqla` / `test-sqla-async` were skipped as dependents. The current head's CI is authoritative.
- Review:
  - Two self-review rounds.
  - An independent Codex review (`gpt-6-sol`, read-only) found a docs inconsistency.
  - Its follow-ups found three older overbroad sentences in the same paragraph, which I corrected.
  - The final whole-paragraph follow-up is CLEAN at `62815d1e`.
  - Records are inline.


pyathena/sqlalchemy/base.py
```diff
@@ -300,8 +300,24 @@ def _get_table(self, connection, table_name: str, schema: str | None = None, **k
...
+        try:
+            with raw_connection.driver_connection.cursor() as cursor:  # type: ignore[union-attr]
+                metadata = self._lookup_table(cursor, schema, name, table_name)
+        except pyathena.error.OperationalError as e:
+            # A federated catalog reports a missing table in its connector's own
+            # words, so ask information_schema whether it exists, as column
+            # reflection does. Throttling still propagates: the query cannot
+            # supply a comment or options, only absence.
+            code = _get_error_code(e.__cause__ or e, unwrap_metadata=True)
+            if not self._is_connector_error(code, catalog):
```
laughingman7743: **Self-review round one (behavior and implementation)**: FINDINGS, repaired

Base `ad1232b2eb19c67c8793a3f86ea7cbdd52366a47`, head `c1dc23d3dd90219b1c5d52118439e3f624274bf6`. This covers `_get_table()`, `_is_fallback_error()`/`_is_connector_error()`, `_get_columns()` as a caller of the shared predicate, the async adapter (it shares `_columns_from_information_schema()` through `_internal_cursor()`), the new tests, and `docs/sqlalchemy.md`. Passes: `/code-review` (high) and `/simplify` (reuse, simplification, efficiency, altitude).

Repaired:
- `pyathena/sqlalchemy/base.py:319`: an earlier revision cached the columns a failed call found. A transient connector error would then have switched later `get_columns()` in that Inspector to `information_schema` without asking the API again. Reverted; the test asserts `info_cache == {}` after every failed call.
- `pyathena/sqlalchemy/base.py:312`: the condition used `code != "MetadataException" or not _is_fallback_error()` to cancel the throttling branch. It now calls `_is_connector_error()`, which `_is_fallback_error()` also uses, so the boundary is defined once.
- `docs/sqlalchemy.md:111-114`: the new sentence contradicted the one before it ("propagate the error"). Merged into one statement with the exception.
- `tests/pyathena/sqlalchemy/test_base.py:339`: the folded test asserted `__cause__ is error` for `NoSuchTableError` too. It fails because the `ClientError` sits one link deeper; fixed after the local run.

Kept, with reasons:
- `catalog_name` values that are Glue-backed but not named `AwsDataCatalog` (`s3tablescatalog/*`, cross-account GLUE catalogs, an explicit `None`) are treated as connectors. This is the `_is_fallback_error()` boundary from #798, unchanged here; the connection default is `"awsdatacatalog"`.
- A failing `information_schema` query replaces the metadata error. `get_columns()` does the same since #798.
- Absent tables are not cached, and each table-level call without cached columns runs one query. Both match the documented rule that absence is not cached.
- The test helper duplicates the inline fakes of pre-existing tests; refactoring those is out of scope.

docs/sqlalchemy.md
```diff
@@ -108,7 +108,10 @@ Other error codes listed in the connection's `RetryConfig.exceptions` are still
 A `retry_config` in `cursor_kwargs` replaces that policy entirely, including those retries, which then run before the fallback.
 The fallback maps unbounded `varchar` to SQLAlchemy `String`, matching Hive `STRING` reflection from the metadata API, and preserves explicit `VARCHAR(n)` and `CHAR(n)` lengths.
 Partition columns are marked from the `extra_info` column.
-This fallback does not populate the table-metadata cache, and table comments and table options still come from the metadata API with the configured retries, so they propagate the error.
+This fallback does not populate the table-metadata cache.
+Table comments and table options still come from the metadata API with the configured retries.
+When that request fails, they propagate the error, with one exception: for an unrecognized `MetadataException` outside `AwsDataCatalog`, they query `information_schema.columns` and raise `NoSuchTableError` if it has no row for the table.
```
laughingman7743: **Self-review round two (claims, callers, operations)**: CLEAN after corrections

Base `ad1232b2eb19c67c8793a3f86ea7cbdd52366a47`, head `c1dc23d3dd90219b1c5d52118439e3f624274bf6`.

Claims checked:
- The Lambda connector message and "`information_schema` returns 0 rows under Lake Formation in `AwsDataCatalog`" are the #786 handoff facts measured in #798.
- "`get_columns()`/`has_table()` already answer through `information_schema`": `_get_columns()` at `base.py:388-399`.
- The PR's Glue statements were measured live on 2026-09-23:
  - `glue:GetTable`, flattened, equals `GetTableMetadata` parameters for Hive DELIMITED/JSON/Parquet, a partitioned table, Glue Iceberg, and S3 Tables.
  - During an Athena throttling episode (30 threads), 1,236 of 1,236 Glue calls succeeded.
- Docs line 111 ("does not populate the table-metadata cache") still holds: this path writes no cache.

Existing callers:
- Signatures, return shapes, and the Glue-catalog behavior are unchanged.
- The only new exception is `NoSuchTableError` for a connector `MetadataException` with no `information_schema` row. It chains the original error.

Operations:
- No retry settings change. `_get_table()` keeps the connection's policy, and the default does not retry `MetadataException`.
- The added query runs only on that connector path: one query, `result_reuse_enable=False`. Throttling adds no query.

Evidence limits:
- No Lambda catalog was available live, so the connector cases are unit-level.
- CI on `36c8126d` failed in `AsyncPandasCursor` tests (empty results). That is unrelated to this path, and its root cause is not established. The current head's CI is pending.

laughingman7743: **Independent review (relayed)**: FINDINGS

Reviewer: Codex CLI 0.156.0, model `gpt-6-sol`, `codex exec --sandbox read-only`, session `01a0ce80-f107-7e51-90f7-c187ca14c9ec`. This is a static review of a detached snapshot. The reviewer ran no tests, had no GitHub access, and did not see the PR text or the author's conclusions.
Base `ad1232b2eb19c67c8793a3f86ea7cbdd52366a47`, head `c1dc23d3dd90219b1c5d52118439e3f624274bf6`. The snapshot was unchanged afterwards.

Covered: the diff, both table-level reflection methods, `_get_columns()` and its cache keys, error unwrapping and chaining, the async cursor adapter, the new tests, and `docs/sqlalchemy.md`.

Finding (Low), quoted: *"`docs/sqlalchemy.md:102`: The page says unrecognized metadata errors propagate, but the new behavior described at line 113 can turn one into `NoSuchTableError`. For example, a federated catalog returning an unrecognized `MetadataException` for a missing table now yields `NoSuchTableError` when `information_schema.columns` returns no rows. The two statements give conflicting guidance."*

Anchored here because `docs/sqlalchemy.md:102` was outside the reviewed diff.

Author verification: confirmed. The sentence was already imprecise for column reflection after #798, and this PR widens the gap.

laughingman7743: **Repair and independent follow-up**: CLEAN at `62815d1e6dd4097c7615ed55a9df45a55f2d45f1`

The repair is docs-only (`docs/sqlalchemy.md`, the absence/failure paragraph). The code and tests are unchanged since `c1dc23d3`.

I rewrote the paragraph so the source of absence is stated explicitly. An error response establishes absence only when it is a recognized `EntityNotFoundException`. An `information_schema` query can then establish absence from its result.

The follow-ups found three overbroad sentences that predate this PR. They were wrong after #777 and #798, and I corrected them because they sit in the paragraph being repaired:
- Line 100: "propagate throttling". Column reflection answers throttling from `information_schema`. The sentence now says a throttling or permission error never *by itself* establishes absence.
- Line 101: `has_table()` "propagates permission failures". A connector can phrase a denial as an unrecognized `MetadataException`. The sentence now says "recognized permission failures".
- Line 114 (new in this PR): the table-level summary omitted the `EntityNotFoundException` → `NoSuchTableError` conversion.

Self-review of each repair:
- Behavior: no code change.
- Claims: each sentence was traced to `_get_table()`, `_get_columns()`, `_is_fallback_error()`, and `_is_connector_error()`.
- `just docs lint`: 0 errors.

Independent follow-ups: Codex CLI 0.156.0, `gpt-6-sol`, `codex exec --sandbox read-only`, a detached snapshot per head, with the range diff from `c1dc23d3`.

| Session | Head | Result |
|---|---|---|
| `01a0ce83-0ed6-7b33-acf8-34489a6371a6` | `d6138cc6` | FINDINGS (line 100) |
| `01a0ce83-fcb3-7f61-9917-3644b306df2a` | `b4f7b9a0` | FINDINGS (line 100 wording) |
| `01a0ce85-1b4c-7450-a0bc-f2e1487fab87` | `4b02892b` | FINDINGS (line 101) |
| `01a0ce86-731f-7093-b30e-c1dc40d9d482` | `74aab19c` | FINDINGS (line 114); whole-paragraph pass |
| `01a0ce87-f61d-7580-901d-f1ec06d4d349` | `62815d1e` | **CLEAN**; every sentence of lines 100–117 traced to the code |

All of these are static reviews. The reviewer noted that the Service Quotas sentence (line 117) cannot be verified from source; it comes from the #780 measurements.

pyathena/sqlalchemy/base.py
laughingman7743: Maintainer feedback applied in `5e8228c3`: the Lambda connector message is inlined at its single call site in `test_table_level_reflection_of_failed_lookup` instead of being a class constant. This is test-only with no behavior change: `uv run --env-file .env pytest -n 1 tests/pyathena/sqlalchemy/test_base.py -k table_level_reflection` gives 9 passed, and `just lint` passes.

